PR #2, opened 2020-07-17
Readme improvements

lemastero:
Nice library!

I try to run examples from README and to be able to one I needed a small tweak, change
`Things.tupled`
to
`(Things.apply _).tupled`

here:
```
implicit val decoder: json.Decoder[Things] = json.Decoder[(String, Int, Boolean)].map((Things.apply _).tupled)
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -198,7 +198,7 @@ Another usecase is if we want to encode a `case class` as an array of values, ra
 ```scala
 case class Things(s: String, i: Int, b: Boolean)
 object Things {
-  implicit val decoder: json.Decoder[Things] = json.Decoder[(String, Int, Boolean)].map(Things.tupled)
+  implicit val decoder: json.Decoder[Things] = json.Decoder[(String, Int, Boolean)].map((Things.apply _).tupled)
 }
 ```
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -198,7 +198,7 @@ Another usecase is if we want to encode a `case class` as an array of values, ra
 ```scala
 case class Things(s: String, i: Int, b: Boolean)
 object Things {
-  implicit val decoder: json.Decoder[Things] = json.Decoder[(String, Int, Boolean)].map(Things.tupled)
+  implicit val decoder: json.Decoder[Things] = json.Decoder[(String, Int, Boolean)].map((Things.apply _).tupled)
```
fommil: weird, this works for me on 2.13. Are you using 2.12? In that case I think I'd rather document the cleaner syntax... or a more verbose version that works on both.

fommil: oh, I see, it's because `object Things` doesn't do `object Things extends (String, Int, Boolean) => Things`

fommil:
```scala
 implicit val decoder: json.Decoder[Things] =
   json.Decoder[(String, Int, Boolean)].map{ case (p1, p2, p3) => Things(p1, p2, p3) }
```

lemastero: Thanks :)

PR #3, opened 2020-07-17
decode Set, Vector, Seq, Array

lemastero:
Support for other collection types:
* Seq, Vector, Set - based on List
* Array

### DIFF
--- a/src/main/scala/zio/json/decoder.scala
+++ b/src/main/scala/zio/json/decoder.scala
@@ -6,6 +6,8 @@ import scala.util.control.NoStackTrace
 import zio.json.internal._
 import Decoder.JsonError
 
+import scala.reflect.ClassTag
+
 // convenience to match the circe api
 object parser {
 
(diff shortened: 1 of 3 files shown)


fommil:
these are already supported, see

```
  implicit def cbf[T[_], A: Decoder](implicit
    CBF: CanBuildFrom[List[A], A, T[A]]
  ): Decoder[T[A]] = list[A].map(lst => CBF(lst).result())
```

can you remove the instances? we can keep the tests.

That said, there is an argument to having decoders specific to Set and Vector for performance reasons, directly using their builder (not going via List), and I'd be ok to see the PR changed in that way. But be careful: implicit priority matters. You'll need to add SortedSet too and it'll need to be higher priority than Set, falling back to CBF. You might want to create a non-implicit abstraction around ReusableBuilder to avoid copy/pasting the code.

Also, please remove the unrelated cleanup of the `build.sbt`. That kind of change falls under the category of "ZIO coding standardisation" and there will probably be a tonne of things like this that can be done to bring zio-json in line with other zio libraries, once we're done with the core functionality :smile:

src/main/scala/zio/json/decoder.scala
```diff
@@ -228,6 +230,29 @@ object Decoder extends GeneratedTuples with DecoderLowPriority1 with DecoderLowP
...
+  implicit def vector[A](implicit A: Decoder[A]): Decoder[Vector[A]] =
+    list[A].map(_.toVector)
+
+  implicit def seq[A](implicit A: Decoder[A]): Decoder[Seq[A]] =
+    list[A].map(_.toSeq)
+
+  implicit def set[A](implicit A: Decoder[A]): Decoder[Set[A]] =
+    list[A].map(_.toSet)
+
+  implicit def array[A:ClassTag](implicit A: Decoder[A]): Decoder[Array[A]] =
```
fommil: I'm don't think it is in ZIO's interests to support mutable data structures in the public API.

fommil: There is another reason not to support `Array[A]`. Since it is so primitive, it is likely that if somebody wants to support it they will want to support it on a per type basis (in Haskell terms, `OverlappingInstances`). e.g. `Array[Byte]` as base16/58/64, `Array[Char]` as a `String`, etc etc. In which case, it's best left free for the user to define.

lemastero:
Closing - at this point this only interferes with the development of library instead of helping. And I do want this thing to happen :)

src/main/scala/zio/json/decoder.scala
lemastero: Removed. Good point.

PR #4, opened 2020-07-18
Add tests for decode Set Map Vector and Set

lemastero:
Added tests and support for Set, Map, Vector.

It is a bit strange that I needed to add 

```scala
  implicit def seq[A: Decoder]: Decoder[Seq[A]] = new Decoder[Seq[A]] {
    def unsafeDecode(trace: List[JsonError], in: RetractReader): Seq[A] =
      builder(trace, in, new mutable.ListBuffer[A])
  }
```

IMHO `cbf` should be able to handle it. The compilation is ok but tests fail with

```scala
assertion failed: Right(Monster(List())) != Right(Monster(List(5XL, 2XL, XL)))
```

as if sth was wrong with `cbf`

### DIFF
--- a/src/main/scala/zio/json/decoder.scala
+++ b/src/main/scala/zio/json/decoder.scala
@@ -235,6 +235,15 @@ object Decoder extends GeneratedTuples with DecoderLowPriority1 with DecoderLowP
       builder(trace, in, new mutable.ListBuffer[A])
   }
 
+  implicit def vector[A](implicit A: Decoder[A]): Decoder[Vector[A]] =
+    list[A].map(_.toVector)
+
+  implicit def seq[A](implicit A: Decoder[A]): Decoder[Seq[A]] =
+    list[A].map(_.toSeq)
+
+  implicit def set[A](implicit A: Decoder[A]): Decoder[Set[A]] =
+    list[A].map(_.toSet)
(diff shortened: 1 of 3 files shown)


src/main/scala/zio/json/decoder.scala
```diff
@@ -235,6 +235,19 @@ object Decoder extends GeneratedTuples with DecoderLowPriority1 with DecoderLowP
...
 
+  implicit def vector[A: Decoder]: Decoder[Vector[A]] = new Decoder[Vector[A]] {
+    def unsafeDecode(trace: List[JsonError], in: RetractReader): Vector[A] =
+      builder(trace, in, new mutable.ListBuffer[A]).toVector
+  }
+
+  // FIXME compiles without seq but tests fails
+  // assertion failed: Right(Monster(List())) != Right(Monster(List(5XL, 2XL, XL)))
+  // maybe sth is off with cbf ?
+  implicit def seq[A: Decoder]: Decoder[Seq[A]] = new Decoder[Seq[A]] {
```
fommil, line 246: I don't think we'd want `seq` in the decoder, if something is wrong with CBF then it should be fixed.

fommil (approved):
thanks!

PR #20, opened 2020-09-22
Rename annotations

### LINKED ISSUE #16: Rename annotations
A goal of this library is to ensure that a single import can bring everything into scope in a way that is unlikely to suffer from name clashes with other entities in the user's application.

In other words, this should be all that is necessary to use ZIO JSON:

```scala
import zio.json._
```

We are nearly to this goal, however, there are several annotations that need to be renamed: 

1. `field` => `jsonField`
2. `discriminator` => `jsonDiscriminator`
3. `hint` => `jsonHint`
4. `no_extra_fields` => `jsonNoExtraFields`

These annotations are located in the file `macros.scala` inside `/src/main/scala/zio/json`. In addition, usages of these annotations exist inside the test directory (and maybe benchmarks), and there is documentation for these annotations in `README.md` that needs to be updated.

fsvehla:
Closes #16

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,19 +125,19 @@ Almost all of the standard library data types are supported as fields on the cas
 
 ## Configuration
 
-By default the field names of a case class are used as the JSON fields, but it is easy to override this with an annotation `@field`.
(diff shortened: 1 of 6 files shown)


jdegoes:
Awesome! 🙏

PR #63, opened 2020-09-27
decodeJsonTransducer

### LINKED ISSUE #10: Add ZTransducer to convert from stream of Char to stream of A
The `JsonDecode` type class can have a method `decodeJsonTransducer`, which returns a `ZTransducer[Any, Exception, Char, A]`, which decodes character strings to values of type `A`.

fsvehla:
Closes #10

This is more complex than I'd like because of the blocking readers.
Interested to see if you see some opportunity for simplification.

- On a push incoming characters are pushed into a queue
  - a worker repeatedly pull from the queue, decodes elements, and pushes elements
- On each push already decoded elements (from prior pulls) are pushed

Because we don't know how many chars we need to push to yield elements, we don't have back-pressure.
Unless I'm missing something obvious `ZTranscoder.splitLines` suffers from the same issue (Akka Streams require to specify a maximumFrameLength argument).

I've manually formatted the method for readability, but happy to remove the `//format: off` post review.

### DIFF
--- a/zio-json/shared/src/main/scala/zio/json/decoder.scala
+++ b/zio-json/shared/src/main/scala/zio/json/decoder.scala
@@ -7,8 +7,8 @@ import scala.util.control.NoStackTrace
 import zio.blocking._
 import zio.json.JsonDecoder.JsonError
 import zio.json.internal._
-import zio.stream.ZStream
-import zio.{ Chunk, ZIO }
+import zio.stream.{ Take, ZStream, ZTransducer }
+import zio.{ Chunk, IO, Queue, Ref, ZIO, ZManaged }
 
 /**
  * A `JsonDecoder[A]` instance has the ability to decode JSON to values of type `A`, potentially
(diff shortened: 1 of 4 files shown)


fsvehla:
@iravid Interested what you think as well

zio-json/shared/src/main/scala/zio/json/decoder.scala
```diff
@@ -71,6 +71,113 @@ trait JsonDecoder[A] { self =>
...
+                            }
+
+                            unsafeDecode(Nil, jsonReader)
+                          } catch {
+                            case t @ JsonDecoder.UnsafeJson(trace) =>
+                              throw new Exception(JsonError.render(trace))
+                          }
+
+                          runtime.unsafeRun(outQueue.offer(Take.single(nextElem)))
+
```
jdegoes, line 132: Putting single characters on a queue is going to mean this is extremely slow.

Unrelated, can we leverage converting a stream to a reader?

fsvehla: This is not a character, but the decoded element.

fsvehla: > Unrelated, can we leverage converting a stream to a reader?

We use `zio.stream.internal.ZReader` here, as does `ZStream#toReader`

jdegoes: Ah, I see, this makes sense!

jdegoes:
I'm good with merging for now. I think we need a long-term solution to this push/pull mismatch but that may require ZIO Streams work.

jdegoes (approved):

PR #65, opened 2020-09-28
wip commit on #75 implementing encoder and decoder for java time classes

brbrown25:
So this is a rough first crack I do have a couple questions
1. How would you like to handle formatters for LocalDateTime and the like
2. I'm not sure what I'm doing wrong on the decoder side, but most of the added tests fail
3. What are your thoughts on how I structured the tests (i.e. for enumerated values should I just map on the values or keep it explicit like it is currently?

Thanks and I look forward to the feedback!

### DIFF
--- a/zio-json/shared/src/main/scala/zio/json/decoder.scala
+++ b/zio-json/shared/src/main/scala/zio/json/decoder.scala
@@ -297,7 +297,7 @@ object JsonDecoder extends GeneratedTupleDecoders with DecoderLowPriority0 {
     final case class SumType(cons: String)       extends JsonError
   }
 
-  implicit val string: JsonDecoder[String] = new JsonDecoder[String] {
+  implicit def string: JsonDecoder[String] = new JsonDecoder[String] {
     def unsafeDecode(trace: List[JsonError], in: RetractReader): String =
       Lexer.string(trace, in).toString
   }
@@ -315,7 +315,7 @@ object JsonDecoder extends GeneratedTupleDecoders with DecoderLowPriority0 {
   implicit val byte: JsonDecoder[Byte]   = number(Lexer.byte)
   implicit val short: JsonDecoder[Short] = number(Lexer.short)
   implicit val int: JsonDecoder[Int]     = number(Lexer.int)
-  implicit val long: JsonDecoder[Long]   = number(Lexer.long)
+  implicit def long: JsonDecoder[Long]   = number(Lexer.long)
   implicit val bigInteger: JsonDecoder[java.math.BigInteger] = number(
     Lexer.bigInteger
   )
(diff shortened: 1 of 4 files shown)


fsvehla:
Thanks for this work, Brandon!

> How would you like to handle formatters for LocalDateTime and the like

I think we can restrict most of the formatted date to ISO8601 for interoperability usability’s sake, for example I don’t anticipate every receiving a Zone like `America/New_York` over the wire, but have seen multiple variants of ISO8601 with up microsecond precision.

> I'm not sure what I'm doing wrong on the decoder side, but most of the added tests fail

```scala
assert("MONDAY".fromJson[java.time.DayOfWeek])(isRight(equalTo(java.time.DayOfWeek.MONDAY)))
```

`fromJson` requires a encoded JSON string - you’re missing the `"` around your it.

> What are your thoughts on how I structured the tests (i.e. for enumerated values should I just map on the values or keep it explicit like it is currently?

They are very readable! Consider moving them to their own files and import java.time._ there.

zio-json/shared/src/main/scala/zio/json/decoder.scala
```diff
@@ -217,6 +217,23 @@ object JsonDecoder extends GeneratedTupleDecoders with DecoderLowPriority0 {
   implicit val bigDecimal: JsonDecoder[java.math.BigDecimal] = number(
     Lexer.bigDecimal
   )
+  implicit val dayOfWeek: JsonDecoder[java.time.DayOfWeek]           = string.map(java.time.DayOfWeek.valueOf(_))
```
fsvehla: Your tests seem to cover the happy path, it would be good to see what happens if some decoding fails.

I’m mentioning this because you’re using `map` with impure functions that might throw. 
Maybe what you want is `mapOrFail`

brbrown25: Yeah that's a great call on that!

brbrown25:
> `fromJson` requires a encoded JSON string - you’re missing the `"` around your it. 
Haha, yeah that makes perfect sense! I'll work on cleaning this up this morning, thanks for the quick feedback!

- [x] move java tests to their own files
- [x] fix decoder test by properly having strings
- [x] use `mapOrFail` when calling impure functions
- [x] stick with `ISO8601` formatting for all places where a formatter can be used

fsvehla:
What should we do about `Instant`? Other JSON libraries seem to use ISO-8601 to en/decode it

jdegoes:
For `Instant`: I think ISO-8601 is fine but probably a number (`Long`) is fine too (epoch millis).

zio-json/shared/src/main/scala/zio/json/decoder.scala
```diff
@@ -297,7 +297,7 @@ object JsonDecoder extends GeneratedTupleDecoders with DecoderLowPriority0 {
     final case class SumType(cons: String)       extends JsonError
   }
 
-  implicit val string: JsonDecoder[String] = new JsonDecoder[String] {
+  implicit def string: JsonDecoder[String] = new JsonDecoder[String] {
```
fsvehla, line 300: I don’t think this change is necessary?

fsvehla (approved):

zio-json/shared/src/main/scala/zio/json/decoder.scala
brbrown25: Without it I was getting a null pointer exception in the new LowPriority trait I created

Issue #91, opened 2020-10-25
Generated encoder for a case class with None produces invalid json


alvaroc1:
```scala
case class User (name: Option[String], age: Int)
object User { 
  implicit val encoder = DeriveJsonEncoder.gen[User]
}
```
```
scala> User(None, 37).toJson
res7: String = {,"age":37}
```

Note the extra comma at the beginning.

If the fields are re-ordered, the json is fine:
```scala
case class User (age: Int, name: Option[String])
object User { 
  implicit val encoder = DeriveJsonEncoder.gen[User]
}
```
```
scala> User(36, None).toJson
res9: String = {"age":36}
```

fsvehla:
Thanks for the report, @alvaroc1.

alvaroc1:
I'll take a crack at fixing it

fsvehla:
Closed by #92

PR #92, opened 2020-10-25
Fix for #91 - extra comma when starting with an empty option field

alvaroc1:
Unit test and fix for #91


fsvehla:
Looks good. Thank you, and congratulations on your first contribution!

PR #118, opened 2020-11-22
Fix Year encoder

### LINKED ISSUE #117: Cannot round trip encode/decode large years
Once #115 is fixed then small and large years cannot be encoded and then decoded.

For example:
```
Year.of(10000).toJson.fromJson
```

Fails with:
```
Left((10000 is not a valid ISO-8601 format, Text '10000' could not be parsed at index 0))
```

steinybot:
Fixes #117


jdegoes:
@steinybot Thank you!

PR #122, opened 2020-11-24
Make jawn-ast dependency for test only

ghostdogpr:
Changed the `jawn-ast` dependency to test only as I think it was intended to be 😄


softinio (approved):

fsvehla:
Thanks, @ghostdogpr !

PR #133, opened 2020-12-09
Fixes #132:  throws AOOBE

fanf:
Change bad value of `0` for `intial` parameter of `WithRecordingReader`.
Add an argument check in `WithRecordingReader` to avoid that error in the future. 
Add a test case for `orElse` combinator - I'm not sure how to use it without `asInstanceOf[]` given variance and inference in each part, which doesn't seems to be a good practice.
Add a comment in `orElse` to explain that it can be unsafe regarding DOS attack, and show an hand-coded alternative in comment.
Add tests.

Issue #132, opened 2020-12-09
`orElse` decoder combinator throws IndexOutOfBound exception


fanf:
A soos as someone try to use `JsonDecoder.orElse`, we get that stack trace: 

```
java.lang.ArrayIndexOutOfBoundsException: 0
	at zio.json.internal.WithRecordingReader.readChar(readers.scala:185)
	at zio.json.internal.OneCharReader.nextNonWhitespace(readers.scala:32)
	at zio.json.internal.OneCharReader.nextNonWhitespace$(readers.scala:30)
	at zio.json.internal.WithRecordingReader.nextNonWhitespace(readers.scala:158)
	at zio.json.internal.Lexer$.char(lexer.scala:283)
	at zio.json.internal.Lexer$.string(lexer.scala:140)
	at zio.json.JsonDecoder$$anon$4.unsafeDecode(decoder.scala:303)
	at zio.json.JsonDecoder$$anon$4.unsafeDecode(decoder.scala:301)
	at zio.json.JsonDecoder$$anon$2.unsafeDecode(decoder.scala:217)
	at zio.json.JsonDecoder$$anon$1.unsafeDecode(decoder.scala:191)
	at zio.json.JsonDecoder.decodeJson(decoder.scala:51)
	at zio.json.JsonDecoder.decodeJson$(decoder.scala:50)
	at zio.json.JsonDecoder$$anon$1.decodeJson(decoder.scala:187)
	at zio.json.package$DecoderOps$.fromJson$extension(package.scala:23)
```

The cause is that `orElse` use `WithRecordingReader` with an initial value of 0.

PR #136, opened 2020-12-15
Add a manualSumSwitch method for hand-coded decoder

fanf:
This PR add a `manualSumSwitch[A]` that can efficiently help in decoding sum type. The typical use case is when the JSON format is fixed and it is missing type information (case class name or field with type). 

Example: 
```
JsonDecoder.manualSumSwitch[AnyVal] {
  case 't' | 'f' => JsonDecoder[Boolean].widen
  case c         => JsonDecoder[Int].widen
}
```

This PR is expecting to have #133 because it provides a better solution for the `orElse` comment added in it and it relies on `.widen` in all of its use cases.

### DIFF
--- a/zio-json/shared/src/main/scala/zio/json/decoder.scala
+++ b/zio-json/shared/src/main/scala/zio/json/decoder.scala
@@ -194,7 +194,7 @@ trait JsonDecoder[A] { self =>
    * alternative would look like:
    *
    * ```
-   * val decoder: JsonDecoder[AnyVal] = JsonDecoder.manualSumSwitch[AnyVal] {
+   * val decoder: JsonDecoder[AnyVal] = JsonDecoder.peekChar[AnyVal] {
    *   case 't' | 'f' => JsonDecoder[Boolean].widen
    *   case c         => JsonDecoder[Int].widen
    * }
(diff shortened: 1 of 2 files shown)


zio-json/shared/src/main/scala/zio/json/decoder.scala
```diff
@@ -298,6 +315,18 @@ object JsonDecoder extends GeneratedTupleDecoders with DecoderLowPriority0 {
     final case class SumType(cons: String)       extends JsonError
   }
 
+  def manualSumSwitch[A](partialFunction: PartialFunction[Char, JsonDecoder[A]]): JsonDecoder[A] = new JsonDecoder[A] {
```
jdegoes: This should be defined in the companion object of `JsonDecoder`. I would suggest calling it `peekChar`.

jdegoes (commented):
One minor change but otherwise looks good!

zio-json/shared/src/main/scala/zio/json/decoder.scala
fanf: I'm not sure I understand the part about the companion object. It's already in object `JsonDecoder`.
I change the name immediatly.

fanf:
I rebased on master to clear out all commits related to PR #132

jdegoes (approved):
Excellent work, thank you! 🙏

PR #143, opened 2020-12-25
Replace CircleCI with GitHub actions

mijicd:
As mentioned in the PR title, I have replaced CircleCI with GitHub actions. However, I noticed a few things in the old CI setup:

- some jobs weren't used (e.g. dotty builds, jmh cross-compilation)
- scala versions weren't compatible with BuildHelper
- JS isn't tested (well, all tests are JVM specific atm)
- Dotty doesn't seem to be working

Please make sure to keep the CI setup tidy and up to date and let me or @softinio know if you need any help.


fsvehla:
@mijicd Those have been on purpose:

- Some jobs weren't used (e.g. dotty builds, jmh cross-compilation)
- JS isn't tested (well, all tests are JVM specific atm)
- Dotty doesn't seem to be working

What do we need to get this over the finishing line?

mijicd:
Well, i have ported the working jmh steps and added the test/compile js step in. Dotty is failing. It's up to you guys to make that one happening.

My goal with this PR is to move away from CircleCI as we're doing that across the org.

fsvehla:
If we can disable the same steps that were disabled with CircleCI we can move right away.

fsvehla:
@mijicd To expand, Dotty and Magnolia is not trivial. It has not worked so far, which is why the steps were disabled.

mijicd:
Yeah, the current setup contains only the passing steps from CircleCI plus the js step :).

That's what i meant with Dotty: once you have it up and running, please add it to actions.

fsvehla:
Understood!

PR #183, opened 2021-02-20
Fix issue with codecs

emarx:
@fommil -- This PR fixes the issue that you and Derander were talking about in Discord the other day. The commit `Red` creates a spec for codecs that mirrors the decoder spec, and introduces a new test, `empty`, that fails the build. The commit `Green` fixes the spec by overriding `unsafeDecodeMissing`. To make the test compile, I create several missing codecs on `JsonCodec` by using the analogous values found on JsonEncoder and JsonDecoder, which should make using `DeriveJsonCodec` much easier in general.

### DIFF
--- a/zio-json/shared/src/main/scala/zio/json/codecs.scala
+++ b/zio-json/shared/src/main/scala/zio/json/codecs.scala
@@ -65,8 +65,8 @@ object JsonCodec {
 
           override def decoder: JsonDecoder[A] = decoder0
 
-//          override def unsafeDecodeMissing(trace: List[JsonError]): A =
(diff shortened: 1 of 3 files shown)


emarx:
I've also created the rest of the low-priority codecs here: https://github.com/emarx/zio-json/pull/2. Also, it would be nice to introduce an `Encoder`, `Decoder`, and `Codec` for Scala's `BigDecimal`, instead of just for the Java version (https://github.com/emarx/zio-json/pull/1). Will submit these after this PR is merged.

fsvehla:
Thanks for the contribution, @emarx!

Issue #181, opened 2021-02-19
case class default field values


dpennell:
It would be useful to support default field values when decoding.

```scala
case class Entity(id: Long, description: String = "")

"""{"id":42}""".fromJson[Entity] == Right(Entity(42L, ""))
```

I can provide a PR if needed.

fsvehla:
>> I can provide a PR if needed.

Go for it!

cartazio:
what release is this in? I'm exactly needing this desperately :)

fsvehla:
Only in snapshot releases so far. The Readme in GitHub has a link to the latest snapshot

cartazio:
@fsvehla thx! (though i'm not finding any way to successfully add the snapshots resolver to my project atm :( )
edit: all the documentation online for snapshot resolver stuff seems to not work for me atm

cartazio:
the error i'm seeing (likely my fault) is 
```
[error]   not found: /home/carter/.ivy2/local/dev.zio/zio-json_2.12/0.1+17-974132cc-SNAPSHOT/ivys/ivy.xml

[error]   not found: https://repo1.maven.org/maven2/dev/zio/zio-json_2.12/0.1+17-974132cc-SNAPSHOT/zio-json_2.12-0.1+17-974132cc-SNAPSHOT.pom
```

cartazio:
derp, i was missing a resolver in one sub build. i'll try it agian :)

fsvehla:
This will be included in a non-snapshot release later this week.

PR #187, opened 2021-02-24
add support for default values for case class fields

dpennell:
If missing in or null, default value will be used.

Do we want an annotation to enable this behavior?

If this looks OK, I'll update the docs.

### DIFF
--- a/docs/overview/decoding.md
+++ b/docs/overview/decoding.md
@@ -34,6 +34,21 @@ Now we can parse JSON into our object
 """{ "curvature": 0.5 }""".fromJson[Banana]
 ```
 
+### Automatic Derivation and case class default field values
+
+If a case class field is defined with a default value and the field is not present or `null`, the default value will be used.
+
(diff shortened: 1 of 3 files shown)


dpennell:
Issue #181

emarx:
This is sweet! IMO seems reasonable to have on by default, and would be great to have a note in the docs about this behavior!

fsvehla:
Looks good! Thank you.

dpennell:
You're welcome, glad I finally was able to contribute.

dpennell:
@fsvehla this appears to have failed to publish the docs after the merge.  It is barfing over some embedded scala code.  I'm not sure how to resolve this.

fsvehla:
@dpennell I’ll check it out

fsvehla:
@dpennell Site fixed. https://zio.github.io/zio-json/docs/overview/overview_decoding#automatic-derivation-and-case-class-default-field-values

PR #202, opened 2021-03-08
Add separate char codec to handle control characters 

thinkharderdev:
Current encoding/decoding is broken for control characters (\t, \n, etc). Need to handle those characters differently for string and single char encoding to be able to successfully decode.


jdegoes:
@fsvehla Can you do a quick review? Would be good to get this in

fsvehla:
Looks good! Thanks, @thinkharderdev.

Will cut a release including this later this week.

Issue #246, opened 2021-04-08
Add Akka http to the list of interop libraries in the documentation


remiguittaut:
Add Akka http to the list of interop libraries in the documentation.

fsvehla:
@remiguittaut You also need to add it to the sidebar: https://zio.github.io/zio-json/docs/interop/interop_index

remiguittaut:
Ha yes, I thought something was off with the menu... Will do.

remiguittaut:
PR done :)

PR #247, opened 2021-04-08
Fixing Issue #246 - Adding Akka Http to the list of interop libraries

remiguittaut:
Fixing Issue #246 - Adding Akka Http to the list of interop libraries in documentation


fsvehla:
Thanks, @remiguittaut!

PR #255, opened 2021-04-19
Publish for scala 3

amitksingh1490:


fsvehla:
@amitksingh1490 Magnolia is not quite ready yet, but very close: https://github.com/propensive/magnolia/issues/246#issuecomment-822263080

PR #265, opened 2021-04-22
Add numeric JsonFieldEncoder / JsonFieldDecoder

fsvehla:
@kitlangton You’ve expressed surprise that we don’t have them — so here they are.


kitlangton:
Wahoo! Thanks :)

PR #267, opened 2021-04-23
JsonCursor[From, To]

fsvehla:
@jdegoes some quick questions here

### DIFF
--- a/zio-json/shared/src/main/scala/zio/json/ast/JsonCursor.scala
+++ b/zio-json/shared/src/main/scala/zio/json/ast/JsonCursor.scala
@@ -4,7 +4,7 @@ import scala.annotation.implicitNotFound
 
 // TODO: Add another type parameter (From) to JsonCursor
 // TODO: Rename to JsonPath[From, To]???
-sealed trait JsonCursor[A] { self =>
+sealed trait JsonCursor[A <: Json] { self =>
   // TODO: Rename to >>>
   def ++[B <: Json](that: JsonCursor[B])(implicit drillDown: JsonCursor.DrillDown[A, B]): JsonCursor[B] = ???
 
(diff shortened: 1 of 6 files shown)


zio-json/shared/src/main/scala/zio/json/ast/ast.scala
```diff
@@ -56,21 +56,30 @@ sealed abstract class Json { self =>
...
         self.get(parent).flatMap { case Arr(elements) =>
           elements.lift(index).map(Right(_)).getOrElse(Left(s"The array does not have index ${index}"))
         }
 
-      case JsonCursor.FilterType(parent, jsonType) =>
-        self.get(parent).flatMap(jsonType.get(_))
+      case JsonCursor.FilterType(parent, t @ jsonType) =>
+        // TODO: When using 'asJson' on jsonType we can't prove that `A' returned from self.get with
+        //       parent of JsonCursor[_$1] == A
+        self.get(parent).flatMap(x => jsonType.get(x))
```
fsvehla, line 103: @jdegoes Using `jsonType.asJson: (A => Json)` here doesn’t work, beause `self.get(parent)` with a parent of `JsonCursor[_]` does not return an `A`, which `jsonType` could turn into `Json`.

zio-json/shared/src/main/scala/zio/json/ast/JsonCursor.scala
```diff
@@ -1,58 +1,60 @@
...
 }
+
 object JsonCursor {
-  @implicitNotFound("Cannot drill down from ${A} into ${B}")
-  sealed trait DrillDown[A, B]
-  object DrillDown {
-    implicit def LeftIsJson[B]: DrillDown[Json, B]    = new DrillDown[Json, B] {}
-    implicit def LeftIsArr[B]: DrillDown[Json.Arr, B] = new DrillDown[Json.Arr, B] {}
-    implicit def LeftIsObj[B]: DrillDown[Json.Obj, B] = new DrillDown[Json.Obj, B] {}
+  implicit class JsonCursorObjOps[From](cursor: JsonCursor[From, Json.Obj]) {
```
fsvehla, line 24: @jdegoes Could not define this on `JsonCursor` without casting, even though the compiler should know that the is sound, this is what it looked like:

```scala
  def field(field: String)(implicit ev: To =:= Json.Obj) =
   JsonCursor.DownField(self.asInstanceOf[JsonCursor[From, Json.Obj]], field
```

fsvehla:
Getting this into the topic branch, we can revisit the PR for discussion if necessary.

zio-json/shared/src/main/scala/zio/json/ast/ast.scala
jdegoes: All right. Well, it was worth a try!

zio-json/shared/src/main/scala/zio/json/ast/JsonCursor.scala
jdegoes: It's because you need to map on the output type. 

In thinking about this, I believe that we should have:

```scala
sealed trait JsonCursor[-From, +To]
```

The use of variance will make composition nicer and eliminate some type casts.

fsvehla: Makes sense, will revisit!

jdegoes: Btw, I always recommend using `A <:< B` instead of `A =:= B` because it's more flexible.

jdegoes: /cc @fsvehla

fsvehla: 👍🏻

PR #268, opened 2021-04-23
Fix #263: Decoder instances for String and Long

### LINKED ISSUE #263: Decoder instances for String and Long
The `Decoder` instances for `String` and `Long` are defined as `def`s.
Shouldn't these be defined as `val`s?

ajaychandran:
Fixes #263,
1. Redefined instances as `val`s.
2. Updated `String` instance dependencies to use `mapStringOrFail`.


fsvehla (approved):
Thanks!

Issue #253, opened 2021-04-15
annotation to hide case class field from encoder


fsvehla:
>> quick question, sorry for not reading about it, but do we have a way in zio-json to hide a field on the case class from being seen on resulting JSON string. Maybe annotation or something..

fsvehla:
Completed in #272

PR #272, opened 2021-04-25
annotation to hide case class field from encoder #253

jgoday:
Define jsonExcludeFields annotation to exclude fields from the encoding JSON.
Added "exclude fields" test case.

`final case class jsonExcludeFields(fields: List[String]) extends Annotation`

### DIFF
--- a/zio-json/shared/src/main/scala/zio/json/macros.scala
+++ b/zio-json/shared/src/main/scala/zio/json/macros.scala
@@ -57,11 +57,6 @@ final case class jsonHint(name: String) extends Annotation
 final class jsonNoExtraFields extends Annotation
 
 /**
- * If used on a case class, will exclude fields from the resulting JSON.
- */
-final case class jsonExcludeFields(fields: List[String]) extends Annotation
-
-/**
  * If used on a case class field, will exclude it from the resulting JSON.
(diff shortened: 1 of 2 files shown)


fsvehla:
@jgoday Could this annotation also be added on the fields? I’m conflicted about the correct approach — just wondering.

jgoday:
> 
> 
> @jgoday Could this annotation also be added on the fields? I’m conflicted about the correct approach — just wondering.

@fsvehla Just updated PR to allow @jsonExclude annotation directly on case class field.
Don't know if both annotations (jsonExcludeFields and jsonExclude) makes sense ? or maybe, it's better to create a more general one and make it work in both cases ...

fsvehla:
@jgoday I think the annotation on the field is sufficient. Please run `prepare` in SBT to fix the lint errors.

jgoday:
> @jgoday I think the annotation on the field is sufficient. Please run `prepare` in SBT to fix the lint errors.

@fsvehla Just removed jsonexcludefields and keep only jsonexclude (as a field annotation). I have executed 'prepare'  too. :+1:

fsvehla:
Thanks of your work on this, @jgoday!

PR #296, opened 2021-05-10
Merge, tests

fsvehla:
Extraction from AST work.

@jdegoes I don’t think that `merge` needs to return `Either`.

At least in my mind in a merge operation the RHS always takes precedence,
so merging a top-level `true` with `null` -> `null`.

The strongest case for merging disparate types is in objects where we could `null`ing a field
when updating it is pretty common. The case for single-node documents is weaker.

LMK what you think.


fsvehla:
Getting this in.

jdegoes:
@fsvehla Looks good to me!

Issue #105, opened 2020-11-11
Support enum encoding


svroonland:
It would be nice to have support for encoding enums as string values, typically `sealed trait`s of only `case objects`.

For example:

```scala
sealed trait StreamStatus
case object UPDATING extends StreamStatus
case object CREATING extends StreamStatus
```

would be encoded as `"UPDATING"` or `"CREATING"`

jgoday:
@svroonland Hi, I have tried to provide a solution with the following PR #307. I think it has a couple of things that can be improved, but I think it is interesting to see if it is an acceptable solution.

svroonland:
@jgoday Cool! I see @fsvehla is already giving some feedback on the PR.

plokhotnyuk:
It works as expected with v0.7.8: https://scastie.scala-lang.org/Acx0Bh9mRg6xRyPVrRoafQ

PR #307, opened 2021-05-15
Support enum encoding

jgoday:
From #105.

It allows to encode/decode enums (as sealed traits and case objects) like
```scala
   sealed trait Color
   case object Green extends Color
   case object Yellow extends Color
```

I think that some things needs some revision or maybe a different/better solution:

1 - When decoding a SealedTrait, lexer tries to consumes '{', if it throws an UnsafeJson exception, we try to match an enum value.
2 - Find a better approach to determine if an object is a case object (should work in scala.js too)
```scala
def isCaseObject(x: Any): Boolean = x.getClass.getName.endsWith("$")
```

### DIFF
--- a/zio-json/shared/src/main/scala/zio/json/macros.scala
+++ b/zio-json/shared/src/main/scala/zio/json/macros.scala
@@ -71,7 +71,17 @@ object DeriveJsonDecoder {
     if (ctx.parameters.isEmpty)
       new JsonDecoder[A] {
         def unsafeDecode(trace: List[JsonError], in: RetractReader): A = {
-          if (no_extra) {
+          // if we expect a case object we check string with his name
+          if (ctx.isObject) {
+            val value = Lexer.string(trace, in).toString
+            if (value == ctx.typeName.short) {
+              ctx.rawConstruct(Nil)
+            } else {
+              throw UnsafeJson(
+                JsonError.Message(s"expected ${ctx.typeName.short} got '$value'") :: trace)
+            }
+          }
+          else if (no_extra) {
             Lexer.char(trace, in, '{')
             Lexer.char(trace, in, '}')
           } else {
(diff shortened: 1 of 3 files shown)


fsvehla:
@fommil I’d appreciate your opinion here.

fommil:
conceptually this is nice but it is too magical for my personal tastes. I'd rather this was its own macro. Firstly it breaks backwards compatibility, and secondly it means that your entire sealed trait encoding can change just because you add a non `case object`.

jgoday:
> conceptually this is nice but it is too magical for my personal tastes. I'd rather this was its own macro. Firstly it breaks backwards compatibility, and secondly it means that your entire sealed trait encoding can change just because you add a non `case object`.

I agree, especially regarding the backwards compatibility and the lexer/exceptions part.
It was just an attempt to implement that use case and experimenting with the existing codebase.
Please, remove the PR if necessary.

fsvehla:
@jgoday If we implement this with another macro, we could keep compatibility and fail compilation if not every member of the sealed hierarchy is a case object — wdyt?

jgoday:
> @jgoday If we implement this with another macro, we could keep compatibility and fail compilation if not every member of the sealed hierarchy is a case object — wdyt?

@fsvehla Sounds good, what should it be called (instead of DeriveJsonDecoder) ?

fsvehla:
@jgoday `DeriveJsonDecoderEnum`?

jgoday:
@fsvehla, I have implemented it like this https://github.com/jgoday/zio-json/commit/2f7160a34e37d60b56bf556796920f255fde39fd

Two different macros (DeriveJsonDecoderEnum/DeriveJsonEncoderEnum) only for the sealed traits/case objects like enums.

Is this a valid approach ? If so, we should implement toJsonAST/fromJsonAST tests and keep running existing annotations with the new encoder/decoder for enums.

PR #320, opened 2021-05-26
Fix #315 by adding missing encoders for collections

### LINKED ISSUE #315: Missing encoder for indexedSeq in v0.1.5
W/A could be adding the following definition:
```
implicit def indexedSeqEncoder[A](implicit encoder: JsonEncoder[A]): JsonEncoder[IndexedSeq[A]] =
  (as: IndexedSeq[A], indent: Option[Int], out: Write) => {
    out.write('[')
    as.foreach {
      var first = true
      a =>
        if (first) first = false
        else if (indent.isEmpty) out.write(',')
        else out.write(", ")
        encoder.unsafeEncode(a, indent, out)
    }
    out.write(']')
  }
```

plokhotnyuk:


fsvehla:
Confused about that one, your new test cases basically dupliate the same assertions just a few lines down -- am I missing something?

plokhotnyuk:
@fsvehla Could you please point on lines of duplicates?

BTW, this PR fixes derivation of codecs for `IndexedSeq` type alias in Scala 2.12 and adds .jvmops to avoid out of memory errors during compilation

fsvehla:
@plokhotnyuk Apologies, I completely misread that. I blame a lack of caffeine.

Thanks for this changeset!

PR #340, opened 2021-06-07
Mergify: Does not use master

sideeffffect:
Btw, have you guys considered using `master`? Mistakes like this are bound to happen when things are different across the repos in our organization.
@fsvehla @mijicd

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -16,7 +16,6 @@ pull_request_rules:
         add: ["type: breaking"]
   - name: merge Scala Steward's PRs
     conditions:
-      - base=master
       - author=scala-steward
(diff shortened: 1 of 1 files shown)


.mergify.yml
```diff
@@ -16,7 +16,6 @@ pull_request_rules:
         add: ["type: breaking"]
   - name: merge Scala Steward's PRs
     conditions:
-      - base=master
```
mijicd, line 19: Should it be changed to `base=develop` ?

sideeffffect: Does it even need to be there? :shrug: It's one less configuration that, as far as I can see, isn't needed and that could potentially be wrong, like in our case here.

mijicd: True, steward triggers it only on default branch.

mijicd (approved):

fsvehla:
>> Btw, have you guys considered using master? Mistakes like this are bound to happen when things are different across the repos in our organization.

This make completely sense because of the common, shared code, we’ll switch to `master`.

PR #365, opened 2021-07-05
Add decoder for java.time.Instant

plokhotnyuk:
+ simplify parsing of negative or more than 4 digit years
+ more precise position in parsing errors


fsvehla:
Lovely work again. Thank you!

PR #367, opened 2021-07-07
Proxy isNothing in JsonCodec.apply

stephendavidmarsh:
`JsonCodec.apply`, which is used to implicitly build a `JsonCodec` from an implicit `JsonEncoder` and `JsonDecoder`, is not currently proxying `isNothing` to the underlying `JsonEncoder`. This results in a `JsonCodec` that does not behave like the `JsonEncoder` it is built from.

PR changes the `isNothing` method to proxy to the underlying `JsonEncoder`.


fsvehla:
Thanks!

PR #375, opened 2021-07-13
Fix serialization of floats and doubles

plokhotnyuk:
The `.` character was inserted without taking in account that its position is greater by 1 for negative numbers.
So the number like `-3.644554028000364E16` was serialized as `-.3644554028000364E16`

Also, stripping of trailing zeros after `.` was not tested properly, so the short number like `1.0E7` was serialized as `1.00E7`.


fsvehla (approved):
Wow, well caught!

PR #376, opened 2021-07-14
Add either combinator to JsonCodec

samuelsayag:
Hello,

JsonCodec combinator that create a `JsonCodec[Either[A,B]` given a `JsonCodec[B]`.
(we added this as part of John's teaching yesterday) 

I ran `sbt compile; test; prepare` as per the Contributor guide.

Please, tell me if something else is to be done.

Thanks :)


samuelsayag:
I have just seen and implicit that shows the combinator has been done already :)

PR #381, opened 2021-07-22
Add eraseEither combinator to JsonEncoder/JsonCodec

samuelsayag:
Hello,

I had a need for this encoder and codec (during a spartan session) and we thought with @jdegoes it could be added to zio.json.

Basically, I had this kind of structure:

```scala
    sealed trait StringOrNumber extends Any
    object StringOrNumber {
      final case class Str(s: String) extends AnyVal with StringOrNumber
      final case class Num(s: BigDecimal)   extends AnyVal with StringOrNumber
}
```
for which I can derive a codec based on an either but the result JSON must be without `Right` or `Left` fields so I needed to code another encoder.

We gave the name `eraseEither` because it's kind of forgetting the either structure to give just the value.

I locally launch `compile` and `test` and prepare `prepare` from the contributor doc it seems all.

### DIFF
--- a/zio-json/shared/src/main/scala/zio/json/encoder.scala
+++ b/zio-json/shared/src/main/scala/zio/json/encoder.scala
@@ -47,6 +47,16 @@ trait JsonEncoder[A] extends JsonEncoderPlatformSpecific[A] {
   final def either[B](that: => JsonEncoder[B]): JsonEncoder[Either[A, B]] = JsonEncoder.either[A, B](self, that)
 
   /**
+   * Returns a new encoder that can accepts an `Either[A, B]` to either, and uses either this
+   * encoder or the specified encoder to encode the two different types of values.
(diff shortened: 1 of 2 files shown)


zio-json/shared/src/main/scala/zio/json/encoder.scala
```diff
@@ -7,6 +7,7 @@ import zio.json.javatime.serializers
 
 import java.util.UUID
 import scala.annotation._
+import scala.annotation.meta.field
```
fsvehla: Superfluous import

samuelsayag: yep thanks. removed it.

samuelsayag:
I put the `eraseEither` in the companion object of the `JsonCodec` it is more practical to use when building a new codec.

PR #394, opened 2021-08-05
Use nested encoder for `isNothing`

kitlangton:
Option's implementation of `isNothing` did not delegate to the nested Encoder in the `Some` case. One could argue for either approach, but I think this makes more sense.


fsvehla (approved):
Agreed

PR #400, opened 2021-08-12
org.http4s -> 0.23.1 cats-effect -> 3.2.2 zio-interop-cats -> 3.1.1.0

alek-w:
org.http4s -> 0.23.1 
cats-effect -> 3.2.2
zio-interop-cats -> 3.1.1.0


alek-w:
@fsvehla Could you tell pls how to run CI tests there ?

PR #401, opened 2021-08-12
Update dependencies

TumppiHA:
http4s-dsl 0.21.25 -> 0.23.1
cats-effect 2.5.3 -> 3.2.2
zio-interop-cats 2.5.1.0 -> 3.1.1.0

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -267,10 +267,10 @@ lazy val zioJsonInteropHttp4s = project
   .settings(
     crossScalaVersions --= Vector("3.0.1"),
     libraryDependencies ++= Seq(
-      "org.http4s"    %% "http4s-dsl"       % "0.21.25",
+      "org.http4s"    %% "http4s-dsl"       % "0.23.1",
       "dev.zio"       %% "zio"              % zioVersion,
-      "org.typelevel" %% "cats-effect"      % "2.5.3",
-      "dev.zio"       %% "zio-interop-cats" % "2.5.1.0"  % "test",
+      "org.typelevel" %% "cats-effect"      % "3.2.2",
+      "dev.zio"       %% "zio-interop-cats" % "3.1.1.0"  % "test",
       "dev.zio"       %% "zio-test"         % zioVersion % "test",
       "dev.zio"       %% "zio-test-sbt"     % zioVersion % "test"
     ),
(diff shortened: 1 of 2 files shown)


alek-w:
@fsvehla could you please release it ?

fsvehla:
@TumppiHA I don’t find HTTP4s 0.23 here: https://http4s.org/versions/ or here: https://http4s.org/changelog/

In what state is it?

TumppiHA:
> @TumppiHA I don’t find HTTP4s 0.23 here: https://http4s.org/versions/ or here: https://http4s.org/changelog/
> 
> In what state is it?

Not sure why they haven't listed 0.22 and 0.23 versions in those webpages. However the latest 0.21 release in https://http4s.org/changelog/ states "The 0.21 series is no longer actively maintained by the team, but we’ll continue to entertain binary compatible patches. All users are still encouraged to upgrade to 0.22 (for Cats-Effect 2) or 0.23 (the latest stable series, on Cats-Effect 3)."

fsvehla:
I wonder if we don’t need two modules then — one for the stable cats-effect-2, and one for the stable cats-effect-3 version of HTTP4s.

Thoughts?

alek-w:
@fsvehla pls have a look:
https://github.com/zio/zio/blob/master/build.sbt#L541
https://github.com/zio/zio/blob/master/build.sbt#L543

I guess it would be easier for zio-json-lib users in case both main-zio-project and zio-json stay on the same version of cats, isn't it?

fsvehla:
ZIO itself doesn’t depend on either cats-effect or HTTP4s, the linked config stanza was for docs (or possible benchmarks, but not core).

I think what needs to be done is creating a module for cats-effect-2 and one for cats-effect-3,
allowing users to update their dependencies at their own accord.

I do have some spare cycles to tackle this, unless you want to go ahead.

TumppiHA:
Go ahead if you have a vision of how this should be done 👍

fsvehla:
In order to reduce maintenance burden we’ll target those versions for 0.2.0 (which will be 0.3.0 for ZIO 2)

- HTTP4S 0.23
- Cats-Effect 3

If there is any interest we’ll consider providing support for HTTP4S 0.22 / Cats-Effect 2 in the future.

Thanks for working on this, @TumppiHA; I’m closing for now and will work on the update in the next few days.

Issue #481, opened 2021-11-08
orElse captures the input forever


vigoo:
I'm parsing potentially huge JSONs with `zio-json` with the goal of extracting only some small amount of metadata without reading the whole thing in memory (where it does not even fit). The source of the JSON is an external system that I don't control; it's encoding requires the use of `orElse`, basically the decoder is `error | huge-payload`. 

The current implementation does not work as expected, instead it (tries to) store the whole input in memory.

I investigated the issue and have a proposed fix for this that I'll describe in the linked issue.

PR #482, opened 2021-11-08
Change behavior of recording readers to not capture more after rewind

vigoo:
This is my proposed fix for https://github.com/zio/zio-json/issues/481

The implementation of `orElse` creates a `WithRecordingReader` and `rewind`s it in case the left decoder fails. The right decoder is called with the same recording reader, so it can read the same input that was read until the left side failed. The reader however continues recording the input even after reaching the end of it's "tape", so, when used on top level as in my case, it essentially records the whole input for no reason.

There are only two places in the code where `WithRecordingReader` and `rewind` is used and I believe in both cases the correct behavior is that we don't continue recording after replayed. This means `rewind` can only be called once. 
Also this made the optimization that the detected nested recording readers unusable so I removed that.

### DIFF
--- a/zio-json/shared/src/main/scala/zio/json/internal/readers.scala
+++ b/zio-json/shared/src/main/scala/zio/json/internal/readers.scala
@@ -148,7 +148,7 @@ private[zio] sealed trait RecordingReader extends RetractReader {
 }
 private[zio] object RecordingReader {
   def apply(in: OneCharReader): RecordingReader =
-    WithRecordingReader(in, 64)
+    new WithRecordingReader(in, 64)
 }
 
 // used to optimise RecordingReader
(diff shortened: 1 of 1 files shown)


vigoo:
It is broken and maybe I was wrong about it being the correct behavior in the other use case too - I'll investigate and if we need both behaviors then restore the original version and just add a new method beside `rewind` to release the capturing that we can use from `orElse` only.

zio-json/shared/src/main/scala/zio/json/internal/readers.scala
```diff
@@ -51,6 +51,12 @@ private[zio] final class UnexpectedEnd
     )
     with NoStackTrace
 
+private[zio] final class RewindTwice
+    extends Exception(
+      "RecordingReader's rewind was called twice"
+    )
+    with NoStackTrace
```
fommil: since this indicates a developer bug, rather than an input error, you might want to keep the stack trace here.

PR #497, opened 2021-11-27
Fixes #496: Referring to non-existent class magnolia...

### LINKED ISSUE #496: Error with Scala.js project during fastLinkJS: Referring to non-existent class magnolia.*
The current build (Scala 3) has an dependency issue with magnolia, which causes the following error output during fastLinkJS for Scala.js projects:

`Referring to non-existent class magnolia...`

The issue is with the following line in build.sbt (should we use `%%%` instead of `%%`):

`"com.softwaremill.magnolia" %% "magnolia-core" % "2.0.0-M9"`

laogao:
Fixes #496: Referring to non-existent class magnolia...

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -71,7 +71,7 @@ lazy val zioJson = crossProject(JSPlatform, JVMPlatform)
     // as per @fommil, optimization slows things down.
     scalacOptions -= "-opt:l:inline",
     scalacOptions -= "-opt-inline-from:zio.internal.**",
-    scalaVersion := ScalaDotty,
+    //scalaVersion := ScalaDotty,
     libraryDependencies ++= Seq(
       "dev.zio"                %%% "zio"                     % zioVersion,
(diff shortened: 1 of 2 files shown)


fsvehla:
@laogao Any idea why the CI build would not have picked up on this failure and how we could reproduce it?

laogao:
@fsvehla This error only shows up in Scala.js projects that use zio-json, not when building zio-json itself. The problem is with the transitive dependency on magnolia-core version 2.0.0-M9. Since build.sbt used `%%` instead of `%%%`, the jar file it found was not the one with .sjsir. When compiling for Scala.js the compiler would complain about missing magnolia classes.

DamianReeves:
Thanks so much for this, I've been banging my head against this issue for a couple of weeks but wasn't skilled enough with Scala.js to be sure it wasn't a stupid user error.

PR #525, opened 2021-12-29
Add xmapOrFail to JsonCodec

AleGlowa:
It adds possibility to define a codec with a decoder that can fail


fsvehla:
Thank you for the contribution, @AleGlowa!

PR #578, opened 2022-02-18
bool

jdegoes:
(cherry picked from commit 8e3f02a9ac58c5951cf2c048808c6bb90e5d9673)

### DIFF
--- a/zio-json/shared/src/test/scala/zio/json/internal/async/DecodingSpec.scala
+++ b/zio-json/shared/src/test/scala/zio/json/internal/async/DecodingSpec.scala
@@ -1,5 +1,6 @@
 package zio.json.internal.async
 
+import zio.json.internal.async.AsyncDecoder.LiteralDecoding
 import zio.test._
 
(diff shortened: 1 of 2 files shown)


zio-json/shared/src/test/scala/zio/json/internal/async/DecodingSpec.scala
```diff
@@ -0,0 +1,91 @@
...
+
+            val leftovers = decoding.feed("tr")
+
+            assertTrue(leftovers == 2) &&
+            assertTrue(registers.registers(0) == ())
+
+            val oneMore = decoding.feed("u")
+
+            assertTrue(oneMore == 1) &&
+            assertTrue(registers.registers(0) == ())
```
fsvehla: Those are unused, or does `assertTrue` raise an exception in ZIO 2 when used with a sync test?

jkobejs: It will not raise it, forgot that is is not used 🤦

PR #585, opened 2022-03-05
add useful json manipulation methods

codingismy11to7:
add methods (mostly copied from circe) that are useful when manipulating JSON

I've found methods like these to be useful in production code, but most recently in unit testing code to start with a good basis and then change portions of json to simulate bad input that could come from clients

happy to make changes if requested, some decisions i made that may not be optimal:

- created traits so as to not blow up the size of `ast.scala`, may be better to not do that for discoverability
- could it be bad that i straight up copied some circe APIs? whole google vs oracle thing
- compiler throws warnings for unused method parameters in `Json`, 
  - could change it to a trait (there's probably some reason for it being an abstract class? and changing it might break binary compatibility? just a guess)
  - could keep warnings
  - could silence warnings
  - could add a bunch of boilerplate and remove default implementations

### DIFF
--- a/zio-json/shared/src/main/scala/zio/json/ast/ast.scala
+++ b/zio-json/shared/src/main/scala/zio/json/ast/ast.scala
@@ -24,6 +24,32 @@ import scala.annotation._
  * JsonValue / Json / JValue
  */
 sealed abstract class Json { self =>
+  def asNull: Option[Unit]         = None
+  def asBoolean: Option[Boolean]   = None
+  def asNumber: Option[Json.Num]   = None
+  def asString: Option[String]     = None
+  def asArray: Option[Chunk[Json]] = None
+  def asObject: Option[Json.Obj]   = None
+
+  def mapBoolean(@nowarn f: Boolean => Boolean): Json                          = self
+  def mapNumber(@nowarn f: java.math.BigDecimal => java.math.BigDecimal): Json = self
+  def mapString(@nowarn f: String => String): Json                             = self
+  def mapArray(@nowarn f: Chunk[Json] => Chunk[Json]): Json                    = self
+  def mapArrayValues(@nowarn f: Json => Json): Json                            = self
+  def mapObject(@nowarn f: Json.Obj => Json.Obj): Json                         = self
+  def mapObjectKeys(@nowarn f: String => String): Json                         = self
+  def mapObjectValues(@nowarn f: Json => Json): Json                           = self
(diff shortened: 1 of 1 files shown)


fsvehla:
> created traits so as to not blow up the size of ast.scala, may be better to not do that for discoverability

Given the relatively small changesets, I don’t we need to extract those traits

> could it be bad that i straight up copied some circe APIs? whole google vs oracle thing

What APIs in particular? `asNumber` / `asArray`? But let’s standardize on `to*` `an*` unless you think the slight difference in termininology matters.

They are completely obvious to me, but CC @jdegoes to make sure.

>  compiler throws warnings for unused method parameters in Json, could change it to a trait (there's probably some reason for it being an abstract class? and changing it might break binary compatibility? just a guess) could keep warnings could silence warnings could add a bunch of boilerplate and remove default implementations

They should not be an issue after inlining.

codingismy11to7:
> Given the relatively small changesets, I don’t we need to extract those traits

changed

> What APIs in particular? `asNumber` / `asArray`? But let’s standardize on `to*` `an*` unless you think the slight difference in termininology matters.

not sure exactly what you mean here; i don't think small terminology differences matter, no

> They should not be an issue after inlining.

seems like it's still throwing warnings (unless you meant something should actually be `@inline`d, I didn't do that):

```
...
[warn] D:\dev\zio-json\zio-json\shared\src\main\scala\zio\json\ast\ast.scala:41:23: parameter value f in method mapObjectValues is never used
[warn]   def mapObjectValues(f: Json => Json): Json                           = self
[warn]                       ^
[warn] D:\dev\zio-json\zio-json\shared\src\main\scala\zio\json\ast\ast.scala:42:24: parameter value f in method mapObjectEntries is never used
[warn]   def mapObjectEntries(f: ((String, Json)) => (String, Json)): Json    = self
[warn]                        ^
[warn] 10 warnings found
```

fsvehla:
What I’ve meant re: Circe is that none of the new methods seem to be non-obvious in a JSON API, so from my POV there is no issue here.

fsvehla:
Please add `@nowarn`, I think we’re good then!

codingismy11to7:
> Please add `@nowarn`, I think we’re good then!

cool, I'll try to figure this out tomorrow, thanks

codingismy11to7:
> Please add `@nowarn`, I think we’re good then!

a decade+ of writing scala and TIL about `@nowarn`. it compiles without warnings now

fsvehla:
Thanks for your work!

PR #594, opened 2022-03-20
Update ZIO from 2.0.0-RC2 to 2.0.0-RC3

earldouglas:
This updates ZIO to 2.0.0-RC3, and replaces `ZManaged` with `Scope`.
For reference, see: <https://github.com/zio/zio/pull/6420>

### DIFF
--- a/zio-json/jvm/src/main/scala/zio/JsonPackagePlatformSpecific.scala
+++ b/zio-json/jvm/src/main/scala/zio/JsonPackagePlatformSpecific.scala
@@ -9,22 +9,22 @@ import java.nio.charset.StandardCharsets
 import java.nio.file.{ Path, Paths }
 
 trait JsonPackagePlatformSpecific {
-  def readJsonAs(file: File): ZStream[Any, Throwable, ast.Json] =
+  def readJsonAs(file: File): ZStream[Scope, Throwable, ast.Json] =
     readJsonLinesAs[ast.Json](file)
 
-  def readJsonAs(path: Path): ZStream[Any, Throwable, ast.Json] =
+  def readJsonAs(path: Path): ZStream[Scope, Throwable, ast.Json] =
     readJsonLinesAs[ast.Json](path)
 
-  def readJsonAs(path: String): ZStream[Any, Throwable, ast.Json] =
+  def readJsonAs(path: String): ZStream[Scope, Throwable, ast.Json] =
     readJsonLinesAs[ast.Json](path)
 
-  def readJsonAs(url: URL): ZStream[Any, Throwable, ast.Json] =
+  def readJsonAs(url: URL): ZStream[Scope, Throwable, ast.Json] =
     readJsonLinesAs[ast.Json](url)
 
-  def readJsonLinesAs[A: JsonDecoder](file: File): ZStream[Any, Throwable, A] =
+  def readJsonLinesAs[A: JsonDecoder](file: File): ZStream[Scope, Throwable, A] =
     readJsonLinesAs(file.toPath)
 
-  def readJsonLinesAs[A: JsonDecoder](path: Path): ZStream[Any, Throwable, A] =
+  def readJsonLinesAs[A: JsonDecoder](path: Path): ZStream[Scope, Throwable, A] =
     ZStream
       .fromPath(path)
       .via(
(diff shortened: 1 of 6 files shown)


zio-json/jvm/src/main/scala/zio/JsonPackagePlatformSpecific.scala
```diff
@@ -9,22 +9,22 @@ import java.nio.charset.StandardCharsets
 import java.nio.file.{ Path, Paths }
 
 trait JsonPackagePlatformSpecific {
-  def readJsonAs(file: File): ZStream[Any, Throwable, ast.Json] =
+  def readJsonAs(file: File): ZStream[Scope, Throwable, ast.Json] =
```
jdegoes: Ordinarily, streams bundle scopes within them, so there is no need to leak `Scope` on the outside. 

The only exception is if you want the resources consumed by the stream to stay open beyond the lifetime of the stream.

What do you think here and elsewhere?

zio-json/jvm/src/main/scala/zio/json/JsonDecoderPlatformSpecific.scala
```diff
@@ -30,10 +30,10 @@ trait JsonDecoderPlatformSpecific[A] { self: JsonDecoder[A] =>
   final def decodeJsonStreamInput[R](
     stream: ZStream[R, Throwable, Byte],
     charset: Charset = StandardCharsets.UTF_8
-  ): ZIO[R, Throwable, A] =
+  ): ZIO[R with Scope, Throwable, A] =
     stream.toInputStream
-      .flatMap(is => ZManaged.fromAutoCloseable(UIO(new java.io.InputStreamReader(is, charset))))
-      .use(readAll)
+      .flatMap(is => ZIO.fromAutoCloseable(ZIO.succeed(new java.io.InputStreamReader(is, charset))))
```
jdegoes: Anytime you use `.use` in the old ZManaged API, it means you need to use `ZIO.scoped { ... }` in the new API, to close the open resources. Right now, they are left open when the function returns.

earldouglas: Would we just wrap the latter part?  e.g.:

```scala
    stream.toInputStream
      .flatMap(is =>
        ZIO.scoped {
          ZIO
            .fromAutoCloseable(ZIO.succeed(new java.io.InputStreamReader(is, charset)))
            .flatMap(readAll)
        }
```

zio-json/jvm/src/main/scala/zio/json/JsonDecoderPlatformSpecific.scala
```diff
@@ -30,10 +30,15 @@ trait JsonDecoderPlatformSpecific[A] { self: JsonDecoder[A] =>
   final def decodeJsonStreamInput[R](
     stream: ZStream[R, Throwable, Byte],
     charset: Charset = StandardCharsets.UTF_8
-  ): ZIO[R, Throwable, A] =
-    stream.toInputStream
-      .flatMap(is => ZManaged.fromAutoCloseable(UIO(new java.io.InputStreamReader(is, charset))))
-      .use(readAll)
+  ): ZIO[R with Scope, Throwable, A] =
```
jdegoes, line 33: This `with Scope` needs to be deleted (!).

zio-json/jvm/src/main/scala/zio/json/JsonDecoderPlatformSpecific.scala
```diff
@@ -43,21 +48,21 @@ trait JsonDecoderPlatformSpecific[A] { self: JsonDecoder[A] =>
    *
    * @see also [[decodeJsonStreamInput]]
    */
-  final def decodeJsonStream[R](stream: ZStream[R, Throwable, Char]): ZIO[R, Throwable, A] =
-    stream.toReader.use(readAll)
+  final def decodeJsonStream[R](stream: ZStream[R, Throwable, Char]): ZIO[R with Scope, Throwable, A] =
```
jdegoes, line 51: This `with Scope` needs to be deleted (!)

zio-json/jvm/src/main/scala/zio/json/JsonDecoderPlatformSpecific.scala
earldouglas: See #601 for fix.

zio-json/jvm/src/main/scala/zio/json/JsonDecoderPlatformSpecific.scala
earldouglas: See #601 for fix.

PR #599, opened 2022-03-24
Monkey-patch support for Scala 3 derives

regiskuckaertz:
I know there is ongoing work to use scala 3 macros or shapely, but in the meantime would you see any problem with this?

Adds support for writing generic derivation a la Scala 3 i.e.
```scala
case class Foo(field: Bar) derives JsonCodec
```


fsvehla:
@regiskuckaertz No, I don’t see an issue here. Looks good to me.

PR #601, opened 2022-03-26
Drop unneeded `with Scope`s

earldouglas:
These are leftover from #594, and unnecessarily constrain the
environment types.


fsvehla:
Apologies, I’ve misunderstood the communication on the original PR.
Will release a new RC today.

PR #621, opened 2022-04-17
Fully qualify json codec macro

kitlangton:
Without this, the user gets an error unless they've imported `zio.json._`


fsvehla (approved):

fsvehla:
Thanks, Kit!

PR #626, opened 2022-04-27
fix json ast object equality

kitlangton:
Ran into a bug today. This should fix it.

### DIFF
--- a/zio-json/shared/src/main/scala/zio/json/ast/ast.scala
+++ b/zio-json/shared/src/main/scala/zio/json/ast/ast.scala
@@ -26,7 +26,7 @@ import scala.annotation._
 /**
  * This AST of JSON is made available so that arbitrary JSON may be included as
  * part of a business object, it is not used as an intermediate representation,
- * unlike most other JSON libraries. It is not advised to `.map` or `.emap`
+ * unlike most other JSON libraries. It is not advised to `.map` or `.mapOrFail`
  * from these decoders, since a higher performance decoder is often available.
  *
  * Beware of the potential for DOS attacks, since an attacker can provide much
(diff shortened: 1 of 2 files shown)


zio-json/shared/src/main/scala/zio/json/ast/ast.scala
```diff
@@ -43,14 +43,12 @@ sealed abstract class Json { self =>
...
-            case Some(l) if l != r => true
-            case _                 => false
-          }
-        }.isEmpty
+      right.forall { case (key, r) =>
+        left.get(key) match {
+          case Some(l) if l == r => true
+          case _                 => false
+        }
+      }
```
coacoas, line 51: If I understand this correctly, this ensures that all values of `right` are present in `left`, but does would still return true if `left` has additional fields not represented in `right`. Is that the expected behavior? In the attached tests, the one way check exists: 
```scala
          val obj1 = Json.Obj(
            "quux" -> Json.Str("1"),
            "bar"  -> Json.Str("2")
          )

          val obj2 = Json.Obj(
            "quux" -> Json.Str("1"),
            "bar"  -> Json.Str("2"),
            "maux" -> Json.Str("3")
          )

          assertTrue(obj1 != obj2)
```
but not the reflective
```scala
          assertTrue(obj2 != obj1)
```

zio-json/shared/src/test/scala/zio/json/ast/JsonSpec.scala
```diff
@@ -44,6 +44,33 @@ object JsonSpec extends ZIOSpecDefault {
...
+            "bar"  -> Json.Str("2")
+          )
+
+          val obj2 = Json.Obj(
+            "quux" -> Json.Str("1"),
+            "bar"  -> Json.Str("2"),
+            "maux" -> Json.Str("3")
+          )
+
+          assertTrue(obj1 != obj2)
```
coacoas: Shouldn't this also check `assertTrue(obj2 != obj1)`?

kitlangton: That'd be a great idea :)

fsvehla (approved):

PR #711, opened 2022-08-20
Fix #705 : Defer unsafeDecodeMissing in JsonDecoder

### LINKED ISSUE #705: JsonDecoder's `orElse` and `map` do not defer `unsafeDecodeMissing` to `self`
- Is this a deliberate decision, or just was not thought of?
- I see there is this `suspend` function, which will at least defer on `unsafeDecodeMissing`, but then as soon as you map that, its lost.
- Also, for example, the decoder for `Option` overrides `map` to defer on `unsafeDecodeMissing`, but if you were to `map` a second time, it would be lost.
- The only reason I can think not to do this would possibly be performance reasons, but that seems like a really trivial thing, and can imagine it will cause way more problems than its worth.

If it is the case that it is desirable to change this behavior, and defer to `self`, I would be interested in making a first contribution to change this.

Kalin-Rudnicki:


zio-json/shared/src/main/scala/zio/json/JsonDecoder.scala
```diff
@@ -340,20 +359,6 @@ object JsonDecoder extends GeneratedTupleDecoders with DecoderLowPriority1 {
...
-          override def unsafeDecodeMissing(trace: List[JsonError]): B =
-            f(None)
-
-          def unsafeDecode(trace: List[JsonError], in: RetractReader): B =
-            f(self.unsafeDecode(trace, in))
-
-          override final def fromJsonAST(json: Json): Either[String, B] =
-            self.fromJsonAST(json).map(f)
-        }
-
```
Kalin-Rudnicki, line 356: This is not needed anymore, as `map` has the correct behavior by default now

zio-json/shared/src/main/scala/zio/json/JsonDecoder.scala
```diff
@@ -129,14 +136,18 @@ trait JsonDecoder[A] extends JsonDecoderPlatformSpecific[A] {
   /**
    * Returns a new codec whose decoded values will be mapped by the specified function.
    */
-  def map[B](f: A => B): JsonDecoder[B] =
+  final def map[B](f: A => B): JsonDecoder[B] =
```
Kalin-Rudnicki, line 139: It seems this was only `non-final` so that the `option` decoder could override it

PR #762, opened 2022-10-18
Support for Scala 3 derives with ZIO 2.x (Forward-port of #599)

guersam:
* Monkey-patch support for Scala 3 derives

* Add expression using json algebra for good measure

Forward-ported the original work by @regiskuckaertz


fsvehla:
Thanks, @guersam!

carlos-verdes:
Which version is this included?
I thought it was working already for 0.3.0

PR #765, opened 2022-10-20
Derived JsonDecoder decodeJson vs fromJsonAST error inconsistency tests

andreisilviudragnea:
```
JsonDecoder {
  final def decodeJson(str: CharSequence): Either[String, A]
  def fromJsonAST(json: Json): Either[String, A]
}
```
- Derived `JsonDecoder`s have different errors for:
  -  missing fields when calling `JsonDecoder.decodeJson` (`".b.c.v2(missing)"`) vs `JsonDecoder.fromJsonAST` (`"Missing fields: v2, v3"`)
  - invalid types when calling `JsonDecoder.decodeJson` (`".b.c.v1(expected '\"' got '1')"`) vs `JsonDecoder.fromJsonAST` (`"Not a string value"`).
- The errors reported by `JsonDecoder.fromJsonAST` seem to be broken because the JSON path to the error is missing from the messages: `"Missing fields: v2, v3"`, `"Not a string value"`.

Scala 3 derived decoders fail at runtime with `StackOverflowError` for now. This might be another bug.

### DIFF
--- a/zio-json/shared/src/test/scala/zio/json/DecoderSpec.scala
+++ b/zio-json/shared/src/test/scala/zio/json/DecoderSpec.scala
@@ -427,6 +427,36 @@ object DecoderSpec extends ZIOSpecDefault {
           assert(decoder.fromJsonAST(jsonString.fromJson[Json].getOrElse(Json.Null)))(
             isLeft(equalTo("Missing fields: v2, v3"))
           )
+        },
+        test("invalid type consistency") {
+          case class C(
+            v1: String
+          )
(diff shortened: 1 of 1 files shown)


andreisilviudragnea:
@fsvehla There seems to be a StackOverflowError problem when running the tests on Scala 3, but I am interested at the moment on the Scala 2 implementation.

Scala 3 decoder generation is different from Scala 2, so I think I might have found another bug or maybe the test is not written in the right file.

PR #769, opened 2022-10-25
Support Scala 3 `derives` syntax

Iltotore:
This pull request adds support for [Scala 3 typeclass derivation syntax](https://docs.scala-lang.org/scala3/reference/contextual/derivation.html).

This allows you (in Scala 3) to replace this sample:
```scala
import zio.json.*
case class User(name: String, age: Int)

given JsonEncoder = DeriveJsonEncoder.gen
given JsonDecoder = DeriveJsonDecoder.gen
```

by
```scala
import zio.json.*
case class User(name: String, age: Int) derives JsonEncoder, JsonDecoder
```

Or using `JsonCodec`:
```scala
import zio.json.*
case class User(name: String, age: Int) derives JsonCodec
```

### DIFF
--- a/zio-json/shared/src/main/scala-3/zio/json/macros.scala
+++ b/zio-json/shared/src/main/scala-3/zio/json/macros.scala
@@ -66,6 +66,11 @@ final class jsonNoExtraFields extends Annotation
  */
 final class jsonExclude extends Annotation
 
+extension (decoderObject: JsonDecoder.type) {
+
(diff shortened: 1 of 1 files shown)


fsvehla:
@Iltotore Would this PR also work for you?

https://github.com/zio/zio-json/pull/762

Iltotore:
@fsvehla it uses a different approach (inheritance instead of extension) but it seems to achieve the same result. Indeed it should also work.

fsvehla:
Closed in favour of https://github.com/zio/zio-json/pull/762

PR #776, opened 2022-11-02
Add Default Parameter for Indent in JsonEncoder#encodeJson

adamgfraser:
Would it make sense to provide a default parameter here so that for the simple case you could just call `jsonEncoder.encodeJson(a)`?


fsvehla:
Will check if this is binary compatible and merge.

PR #781, opened 2022-11-10
fix(macros): fix fromJsonAST derivation for Scala 3

### LINKED ISSUE #780: Option decoder for Scala 3.2
I have the following code:

```scala
import zio._
import zio.json._
import zio.json.ast._
import scala.io.Source

final case class SyncState(
  @jsonField("next_batch") nextBatch: String,
  // @jsonField("account_data") accountData: Option[AccountData] = None
  @jsonField("account_data") accountData: Option[String]
)

object Main extends ZIOAppDefault {
  implicit lazy val syncStateDecoder: JsonDecoder[SyncState] = DeriveJsonDecoder.gen
  
  val content = Source.fromFile("file.json").getLines().mkString
  
  override def run =
    ZIO.fromEither(content.fromJson[Json].flatMap(_.as[SyncState])).debug
    // ZIO.fromEither(content.fromJson[SyncState]).debug
}
```

That I cross compile for 2.12, 2.13 and 3.2, and I get the following output
```bash
[45/129] examples[2.12.17].run 
SyncState(s737036_20146721_28996_716058_326898_115_59067_740531_0,None)
[87/129] examples[2.13.10].run 
SyncState(s737036_20146721_28996_716058_326898_115_59067_740531_0,None)
[129/129] examples[3.2.0].run 
<FAIL> Fail(Missing fields: account_data,Stack trace for thread "zio-fiber-4":
        at <empty>.Main.run(Main.scala:19))
timestamp=2022-11-10T17:45:29.830296Z level=ERROR thread=#zio-fiber-0 message="" cause="Exception in thread "zio-fiber-4" java.lang.String: Missing fields: account_data
        at <empty>.Main.run(Main.scala:19)"
```

Note that I've added the `"-Yretain-trees"` flag to my compiler for Scala 3.
My build file look like that (mill):
```scala
import mill._
import scalalib._
import mill.scalalib._

object Versions {
  val zioVersion     = "2.0.3"
  val zioJsonVersion = "0.3.0"
}

val scalaVersions = List("2.12.17", "2.13.10", "3.2.0")

abstract class ExtendedCrossScalaModule(crossScalaVersion: String) extends CrossScalaModule {

  override def scalacOptions =
    if (crossScalaVersion.startsWith("3."))
      Seq("-Yretain-trees")
    else
      Seq.empty[String]
}

object examples extends Cross[ExamplesModule](scalaVersions: _*)

class ExamplesModule(val crossScalaVersion: String) extends ExtendedCrossScalaModule(crossScalaVersion) {
  import Versions._

  def mainClass               = Some("Main")
  override def finalMainClass = T("Main")

  override def ivyDeps = Agg(
    ivy"dev.zio::zio:${zioVersion}",
    ivy"dev.zio::zio-json:${zioJsonVersion}"
  )
}
```
and the `file.json` that I'm using to reproduce the crash
```json
{
    "next_batch": "s737036_20146721_28996_716058_326898_115_59067_740531_0",
    "device_unused_fallback_key_types": []
}
```

I was not able to reproduce that in my tests, this is only happened (as far as I could say) when running the application. I also tried to package it inside a jar and a docker to make sure that this was not related to mill/the build, but I could still reproduce, I don't know why this is not triggered inside my tests.

If I set a default value to the option (such as None), it works fine (for the three version), I guess I could go with that for now, but there is still something fishy going on here I think.
I could try to have a look at this issue, but I don't really know zio-json source code and would like some pointers on where to start looking for that bug.

I think this might be related to #779 

Edit: I'm in fact (with no surprise) able to reproduce it in my tests:
```scala
import zio._
import zio.test._
import zio.test.Assertion._
import zio.json._
import Main._

object MainSpec extends ZIOSpecDefault {

  def spec = suite("test")(
    test("json") {
      val content = """ 
{
    "next_batch": "s737036_20146721_28996_716058_326898_115_59067_740531_0",
    "device_unused_fallback_key_types": []
}
            """
      assert(content.fromJson[SyncState])(
        isRight(equalTo(SyncState("s737036_20146721_28996_716058_326898_115_59067_740531_0", None)))
      )
    }
  )
}
```

ex0ns:
The fromJsonAST method derived inside the macro from Scala 3 does not perform the same check regarding Options without default value than it does for Scala 2.12 and 2.13.

This aims to fix that by simply adding the existing change inside the Scala 3 macros.

Closes #780

### DIFF
--- a/zio-json/shared/src/main/scala-3/zio/json/macros.scala
+++ b/zio-json/shared/src/main/scala-3/zio/json/macros.scala
@@ -195,7 +195,14 @@ object DeriveJsonDecoder extends Derivation[JsonDecoder] { self =>
                   if (defaults(i).isDefined) {
                     ps(i) = defaults(i).get
                   } else {
-                    missing.add(names(i))
+                    val tc = tcs(i)
+                    try {
+                      // There is no default value, see if the type class decoder can handle it:
+                      ps(i) = tc.unsafeDecodeMissing(JsonError.ObjectAccess(names(i)) :: Nil)
+                    } catch {
+                      case JsonDecoder.UnsafeJson(trace) =>
+                        missing.add(names(i))
+                    }
                   }
                 }
                 i += 1
--- a/zio-json/shared/src/test/scala/zio/json/DecoderSpec.scala
+++ b/zio-json/shared/src/test/scala/zio/json/DecoderSpec.scala
@@ -77,6 +77,15 @@ object DecoderSpec extends ZIOSpecDefault {
           assert("""{ "id": 1, "opt": 42 }""".fromJson[WithOpt])(isRight(equalTo(WithOpt(1, Some(42))))) &&
           assert("""{ "id": 1 }""".fromJson[WithOpt])(isRight(equalTo(WithOpt(1, None))))
         },
+        test("option - fromJsonAST") {
+          case class WithOpt(id: Int, opt: Option[Int])
+          implicit val decoder: JsonDecoder[WithOpt] = DeriveJsonDecoder.gen
+
+          assert("""{ "id": 1, "opt": 42 }""".fromJson[Json].flatMap(decoder.fromJsonAST))(
+            isRight(equalTo(WithOpt(1, Some(42))))
+          ) &&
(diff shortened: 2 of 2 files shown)


fsvehla:
Very nice. Thanks.

ex0ns:
You are welcome, thanks for the fast reply (finding the issue has taken more time that fixing it, as the code was already written)

PR #814, opened 2022-12-05
fix snake case

senia-psm:
Fixing `SnakeCase` and `KebabCase` formatters. Snake case of `abc123Def` should be `abc123_def`, not `abc_123_def`.

Keeping old formatters in `ziojson_03` object.

### DIFF
--- a/zio-json/shared/src/main/scala-2.x/zio/json/macros.scala
+++ b/zio-json/shared/src/main/scala-2.x/zio/json/macros.scala
@@ -50,6 +50,7 @@ case object CamelCase extends JsonMemberFormat {
   override def apply(memberName: String): String =
     jsonMemberNames.enforceCamelOrPascalCase(memberName, toPascal = false)
 }
+
 case object PascalCase extends JsonMemberFormat {
   override def apply(memberName: String): String = jsonMemberNames.enforceCamelOrPascalCase(memberName, toPascal = true)
 }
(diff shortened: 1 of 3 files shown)


vigoo (approved):

plokhotnyuk:
@vigoo Why it is merged without proper review and discussion? 

According to [Rust](https://doc.rust-lang.org/1.0.0/style/style/naming/README.html) naming conventions numbers should be separated, while [Ruby](https://github.com/rubocop/ruby-style-guide#snake-case-symbols-methods-vars-with-numbers) has an opposite preference in their style guide.

vigoo:
@plokhotnyuk sorry for that, the solution was discussed by me and @senia-psm on other channels.

I asked him to keep the previous implementation available for supporting format  and reviewed the resulting PR. The new style was selected as default because it's consistent with tapir and zio-config's convention.

senia-psm:
Common use case is `tapir`. If one have tapir schema with `Configuration.default.withSnakeCaseMemberNames` and `zio-json` with `SnakeCase` it will result in inconsistent encoding, which is quite difficult to catch in tests.

plokhotnyuk:
The idea of `enforce...` functions was to accept all types of conventions and output selected one.

But currently both new and old functions don't enforce proper separation/joining of non-alphabetic (including numbers) parts.
As an example:
- the old function fails a conversion of `O-OOO111` to `o_ooo_111`
- the new function fails a conversion of `O-OOO-111` to `o_ooo111` or `o-ooo-111` to `o_ooo111`

plokhotnyuk:
A fix for the lower-cased input could be replacing of 
```
if (ch == '_' || ch == '-') {
  sb.append(separator)
  false
```
by
```
if (ch == '_' || ch == '-') {
  if (i > 1 && i < len && !isAlphabetic(s.charAt(i))) isPrecedingNotUpperCased
  else {
    sb.append(separator)
    false
  }
```

senia-psm:
I don't believe we should convert `O-OOO-111` to `o_ooo111`. Unlike `ooo111Xyz`, in `O-OOO-111`  `OOO` and `111` are different parts from the beginning, it's not up to interpretation.

senia-psm:
i.e. in kebab-case we know if number is an independent part or not: `at-least-2` vs `version2-factory`. In camelCase (`atLeast2` and `version2Factory`) we just don't have this information and we have to select 1 interpretation.

Issue #864, opened 2023-01-22
Feature request: field name aliases


erikvanoosten:
To support evolution of data structures it would be nice if zio-json supports aliases on case class field names, e.g. with an annotation like:

```scala
case class Meeting(@jsonAliases("room") location: String)
```

The idea is that during encoding `location` is used, during decoding both `location` and `room` are used. When both are present decoding results in an error.

Ideally, the annotation supports multiple aliases.

This feature would be similar to Avro's aliases and Jackson's `@JsonAlias` annotation.

erikvanoosten:
Fixed in #866.

PR #866, opened 2023-01-24
Aliases

erikvanoosten:
Add support for aliases by defining an annotation.
Solves #864.


fsvehla:
This looks promising. Can you lint the code so that we can see whether tests pass?

erikvanoosten:
@fsvehla The tests pass now. I will add some unit tests to see how the new feature works. I am particularly interested to see what error messages will look like.

erikvanoosten:
@fsvehla Can you take another look? As far as I can see it is complete now (except for Scala 3).

fsvehla:
Is Scala 3 support something you can tackle?

erikvanoosten:
> Is Scala 3 support something you can tackle?

Sure. Just not today anymore.
BTW, lets not forget documentation :)

erikvanoosten:
@fsvehla I can't get the scala 3 tests to pass on my machine (tried with `sbt ++3.2.1 testJVM`). It seems as if sbt doesn't pick up my changes to the scala 3 macros at all. Any idea what I am doing wrong?

The StringMatrix unit tests now take quite a long time (~150 seconds on my machine). I think that the complex generator expression is to blame.

I wanted to add a unit test that asserts that code with an alias containing the name of an already defined field fails to compile. Something like:

```
assertFailsCompilation {
  case class Person(@jsonAliases("n") name: String, n: Int)
}
```
Any idea on how to do achieve this?

Otherwise, I think this feature is good to go. Any comments and feedback is welcome.

erikvanoosten:
I see some more tests are failing. Any help to fix those is appreciated.

erikvanoosten:
In the end it was a silly mistake. The scala 3 tests pass now.

I would still like to introduce a test like:

```scala
"case class Person(@jsonAliases("n") name: String, n: Int)" shouldNot compile
```

I tried to use `typeCheck` but it doesn't work as I expected (see commented out tests).
We could also try scalatest's support (as in the example above). Is it okay if I pull in scalatest for this?

erikvanoosten:
@fsvehla could you start the build again please?

erikvanoosten:
> We could also try scalatest's support (as in the example above). Is it okay if I pull in scalatest for this?

I suspect that Magnolia doesn't actually evaluate the `join` method when the `gen` method is invoked. If that it is true, I do not see how we can improve this PR any further (except by removing the commented out tests).

erikvanoosten:
@fsvehla All cleaned up and ready for a full review and merging...

erikvanoosten:
@fsvehla Sorry, I forgot to run the tests for scala 3 locally. Should be fixed now.

fsvehla (dismissed):

erikvanoosten:
Its not fixed though. Scala 3 uses different class names then scala 2. I'll have to be smart with the test assertions...  I will fix it tomorrow.

erikvanoosten:
Okay, should be fine now for Scala 2 and 3 on the JVM. Let's see how all the different versions and platforms do...

erikvanoosten:
Aaarrggh, now I forgot the formatting.
Fixed!

All other jobs are green! :)

erikvanoosten:
All green! 😄

fsvehla:
Great work, thanks!

PR #893, opened 2023-03-03
Add ability to decode json in different ways and combine the results

### LINKED ISSUE #487: Behavior of zip
After using Circe, I expected `zip` to behave like `product`:

```scala
import zio.json._

case class A(a: Int)
object A {
    implicit val codec: JsonCodec[A] = DeriveJsonCodec.gen
}

case class B(b: Int)
object B {
    implicit val codec: JsonCodec[B] = DeriveJsonCodec.gen
}

implicit val product: JsonDecoder[(A, B)] = JsonDecoder[A] <*> JsonDecoder[B]

"""
{"a":1,"b":2}
""".fromJson[(A, B)]
```

However, the implementation expects an array. This makes sense as a default, but how can I define the behavior above?

pjoneswork:
This is a different zip functionality. Instead of combining decoders with inside a tuple, this allows us to decode the same json twice. 

Given the setup:

```scala
final case class Foo(a: Int)
final case class Bar(b: String)

implicit val fooDecoder: JsonDecoder[Foo] = DeriveJsonDecoder.gen[Foo]
implicit val barDecoder: JsonDecoder[Bar] = DeriveJsonDecoder.gen[Bar]
```

The old `zip` does

```scala
implicit val fooAndBarDecoder: JsonDecoder[(Foo, Bar)] = JsonDecoder[Foo].zip(JsonDecoder[Bar])
val json = """[{"a": 1}, {"b": "2"}]"""
val result: (Foo, Bar) = json.fromJson[(Foo, Bar)]
```

while `newZip` does

```scala
implicit val fooAndBarDecoder: JsonDecoder[(Foo, Bar)] = JsonDecoder[Foo].newZip(JsonDecoder[Bar])

val json = """{"a": 1, "b": "2"}"""
val result: (Foo, Bar) = json.fromJson[(Foo, Bar)]
```

Fixes zio/zio-json#487

### DIFF
--- a/zio-json/shared/src/main/scala/zio/json/JsonDecoder.scala
+++ b/zio-json/shared/src/main/scala/zio/json/JsonDecoder.scala
@@ -201,6 +201,37 @@ trait JsonDecoder[A] extends JsonDecoderPlatformSpecific[A] {
   final def zipWith[B, C](that: => JsonDecoder[B])(f: (A, B) => C): JsonDecoder[C] =
     self.zip(that).map(f.tupled)
 
+  /**
+   * Returns a new codec that combines this codec and the specified codec into a single codec that
+   * decodes the input into a tuple of the values by the respective codecs.
+   */
+  final def newZip[B](that: => JsonDecoder[B]): JsonDecoder[(A, B)] = new JsonDecoder[(A, B)] {
+      def unsafeDecode(trace: List[JsonError], in: RetractReader): (A, B) = {
+        val in2 = new zio.json.internal.WithRecordingReader(in, 64)
+        val a = self.unsafeDecode(trace, in2)
+        in2.rewind()
+        val b = that.unsafeDecode(trace, in2)
+        (a, b)
+      }
+  }
+
+  /**
+   * NewZips two codecs, but discards the output on the right hand side.
+   */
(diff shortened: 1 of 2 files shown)


zio-json/shared/src/main/scala/zio/json/JsonDecoder.scala
```diff
@@ -201,6 +201,37 @@ trait JsonDecoder[A] extends JsonDecoderPlatformSpecific[A] {
   final def zipWith[B, C](that: => JsonDecoder[B])(f: (A, B) => C): JsonDecoder[C] =
     self.zip(that).map(f.tupled)
 
+  /**
+   * Returns a new codec that combines this codec and the specified codec into a single codec that
+   * decodes the input into a tuple of the values by the respective codecs.
+   */
+  final def newZip[B](that: => JsonDecoder[B]): JsonDecoder[(A, B)] = new JsonDecoder[(A, B)] {
```
987Nabil, line 208: I think this is a good feature. But the naming is nod good ;). I think the right name is `and`.
@fsvehla maybe you have an opinion too?

guizmaii: I agree the name is meh. `and` is a good suggestion :)

PR #921, opened 2023-04-03
Refactor write() method for improved readability and maintainability

samadpls:
Rewrote the write() method to improve its structure and readability, and to reduce the amount of mutable state. The changes included:

- Using early returns to simplify the control flow and reduce nesting
- Using a for loop instead of a while loop for more concise code

These changes make the code easier to understand and maintain, and reduce the potential for errors caused by mutable state. The logic for handling JSON formatting has been preserved and should work as before.


jdegoes:
Unfortunately these will have a negative impact on performance.

PR #922, opened 2023-04-04
enabled dotty release for http4s interop

Adriani-Furtado:
I have published and used it locally and it seems like there is no issues


miroslavmachura:
Hi, would like to see this merged, is there anything preventing that?

987Nabil:
@plokhotnyuk can we merge this?

PR #929, opened 2023-04-12
Migrate JsonEncoder/Decoder macro to Scala 3 native macro. No more magnolia

987Nabil:
- annotation handling
- decoders for recursive data structures

### DIFF
--- a/zio-json/shared/src/main/scala/zio/json/internal/readers.scala
+++ b/zio-json/shared/src/main/scala/zio/json/internal/readers.scala
@@ -161,7 +161,7 @@ final class WithRetractReader(in: java.io.Reader) extends RetractReader with Aut
 private[zio] sealed trait RecordingReader extends RetractReader {
   def rewind(): Unit
 }
-private[zio] object RecordingReader {
+object RecordingReader {
   def apply(in: OneCharReader): RecordingReader =
     new WithRecordingReader(in, 64)
 }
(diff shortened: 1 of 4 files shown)


987Nabil:
Still needs some refactoring. But it would be nice, if one could approve the workflows.

zio-json/shared/src/main/scala/zio/json/internal/readers.scala
```diff
@@ -161,7 +161,7 @@ final class WithRetractReader(in: java.io.Reader) extends RetractReader with Aut
 private[zio] sealed trait RecordingReader extends RetractReader {
   def rewind(): Unit
 }
-private[zio] object RecordingReader {
+object RecordingReader {
   def apply(in: OneCharReader): RecordingReader =
```
987Nabil, line 165: The macro basically inlines the code that calls this apply. Therefore we get a compile time error if it is package private.

987Nabil:
I thought one could extract the common macro code to some helper. But this does not seem to work since the imports of the `using Quotes` are path dependent. So I guess this is ready for review.

jdegoes:
@987Nabil Looks great! Can you add some more tests for this under Scala 3 so we can be more assured the derivation is working correctly?

987Nabil:
There are actually a lot of shared tests. I think most cases are covered. But I can take a look if I find something missing.
Do you have something particular in mind?

plokhotnyuk:
@987Nabil Using [this](https://github.com/plokhotnyuk/jsoniter-scala/tree/master/jsoniter-scala-benchmark) benchmark set you can test both correctness and performance of the new way of derivation on ~60 synthetic and real-world JSON samples.

987Nabil:
fyi: The benchmark does not compile. Did not find the time yet to take a deeper look.

PR #932, opened 2023-04-15
Port implicit macro config to Scala 3

mvv:


mvv:
Transient error in build cache restoration

mvv:
Fixes #943

PR #941, opened 2023-05-09
yaml encoding didn't work for long sized integers.

josephhajduk:
changed the Json.Num case in jsonToYaml to use BigDecimal.longValue instead of intValue

added a test

contributing link in the readme is a[dead link](https://zio.dev/about/contributing) so not sure if you need anything else


987Nabil:
@plokhotnyuk I think we can merge this?

PR #1004, opened 2023-08-22
Fixes #1003: correct indent of seq of objects

### LINKED ISSUE #1003: Default Yaml configuration leads to incorrect Yaml
The current configuration options for SnakeYaml layer leads to it outputing badly formatted yaml like: 

```
id: foo
version: 42
params:
  -
  id: p1
  mandatory: true
```

It seems that `setIndentWithIndicator(true)` option is missing (and not exposed). 

See https://gist.github.com/fanf/445e9228cd62bb786960ce2242c1020c for more details.

fanf:
This PR addresses #1003 as shown in the added test. The problem was just linked to the default value of `IndentWithIndicator`. 

To avoid having an other case where an user can't change a non-exposed config parameter of SnakeYaml, I also added in `YamlOption` the way to get a new base `DumpsterOptions` so that one can always pre-set it with what is needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ The goal of this project is to create the best all-round JSON library for Scala:
 In order to use this library, we need to add the following line in our `build.sbt` file:
 
 ```scala
-libraryDependencies += "dev.zio" %% "zio-json" % "0.6.0"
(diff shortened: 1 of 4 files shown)


vigoo (approved):
Good to go if CI passes

fsvehla (approved):

PR #1011, opened 2023-08-28
Change `zio.json.ast.Json` from sealed abstract class to sealed trait

987Nabil:
There is no benefit, in having here an abstract class instead of a trait, as far as I can tell.
But for derivation in the zio-ecosystem, typically abstract class is not supported (zio-schema for example).
That's why I propose to change to sealed trait. The behaviour should not change at all in my opinion.

### DIFF
--- a/zio-json/shared/src/main/scala/zio/json/ast/ast.scala
+++ b/zio-json/shared/src/main/scala/zio/json/ast/ast.scala
@@ -38,7 +38,7 @@ import scala.annotation._
  * excessive amount of heap memory).
  * JsonValue / Json / JValue
  */
-sealed abstract class Json { self =>
+sealed trait Json { self =>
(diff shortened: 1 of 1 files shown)


guizmaii:
IIRC, there's a bin compat issue with sealed trait. IIRC, @fommil knows that very well as it's in his book?

fsvehla:
Yes, I think it would be an issue; we’d need to push 0.7

fommil:
shouldn't the bug report be with other projects, that didn't consider this well-known pattern? This is like complaining that you want coke to change the shape of their bottle because it doesn't fit in your mini-fridge.

987Nabil:
I found another/better solution for this. I realised that deriving the schema won't work anyway. I'll add a custom schema to the json module of zio-schema.

PR #1043, opened 2023-11-03
Moved the message about scalac -Yretain-trees

lachezar:
Now it appears in the "Automatic Derivation and case class default field values" as well


lachezar:
There was a question about this in the Discord and I've decided to add it in the relevant documentation.

guizmaii (approved):

PR #1054, opened 2024-01-02
Fixes jsonExclude annotation for Scala 3

alxdm:
It should fix the case when `@jsonExclude` field is placed before necessary fields.
The test also was modified in order to reproduce the case.


fsvehla:
Thank you for the contribution!

PR #1068, opened 2024-02-21
Feat: scala3 enumeration support

ThijsBroersen:
With this PR I have tried to implement Scala 3 enumeration support. Because the current macros already do some assumptions and are build with Magnolia I did not want to disrupt to much and tried to implement it under the following condition:
If all subtypes are named values (objects without parameters) then the type is an enumeration.

This means that if an enum has mixed children, with and without parameters, then the implementations stays the same. Only for pure enumerations (all named values) the encoding/decoding will be affected. 

Breaking:
To my opinion the current implementation is wrong as enumerations are normally serialized as strings. But this PR will of course be breaking for the few using zio-json with Scala 3 enumerations already. 

I deliberately did not try to implement this for Scala 2 as these are not native enumerations.

### DIFF
--- a/zio-json/shared/src/main/scala-3/zio/json/macros.scala
+++ b/zio-json/shared/src/main/scala-3/zio/json/macros.scala
@@ -402,7 +402,7 @@ object DeriveJsonDecoder extends Derivation[JsonDecoder] { self =>
     lazy val namesMap: Map[String, Int] =
       names.zipWithIndex.toMap
 
-    def isEnumeration = ctx.subtypes.forall(_.typeclass.isInstanceOf[CaseObjectDecoder[?, ?]])
+    def isEnumeration = ctx.isEnum && ctx.subtypes.forall(_.typeclass.isInstanceOf[CaseObjectDecoder[?, ?]])
 
     def discrim = ctx.annotations.collectFirst { case jsonDiscriminator(n) => n }
 
@@ -641,7 +641,7 @@ object DeriveJsonEncoder extends Derivation[JsonEncoder] { self =>
   def split[A](ctx: SealedTrait[JsonEncoder, A]): JsonEncoder[A] = {
     val jsonHintFormat: JsonMemberFormat =
       ctx.annotations.collectFirst { case jsonHintNames(format) => format }.getOrElse(IdentityFormat)
-    val isEnumeration = ctx.subtypes.forall(_.typeclass == caseObjectEncoder)
+    val isEnumeration = ctx.isEnum && ctx.subtypes.forall(_.typeclass == caseObjectEncoder)
 
     val discrim = ctx
       .annotations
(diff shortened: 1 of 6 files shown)


zio-json/shared/src/main/scala-3/zio/json/macros.scala
```diff
@@ -490,16 +515,18 @@ object DeriveJsonDecoder extends Derivation[JsonDecoder] { self =>
   }
 }
 
+private lazy val caseObjectEncoder = new JsonEncoder[Any] {
```
ThijsBroersen, line 537: This static encoder make it possible to filter on at L625

zio-json/shared/src/main/scala-3/zio/json/macros.scala
```diff
@@ -595,13 +622,43 @@ object DeriveJsonEncoder extends Derivation[JsonEncoder] { self =>
     }
 
   def split[A](ctx: SealedTrait[JsonEncoder, A]): JsonEncoder[A] = {
+    val isEnumeration = ctx.subtypes.forall(_.typeclass == caseObjectEncoder)
```
ThijsBroersen: I chose to use this hack because I don't think I have access to the `ctx.params` of the subtypes here anymore.

ThijsBroersen:
Interesting, so it seems shared specs are now failing because in Scala 3 the sealed traits are now also treated as enumerations. 🤔

ThijsBroersen:
> Interesting, so it seems shared specs are now failing because in Scala 3 the sealed traits are now also treated as enumerations. 🤔

Fixed 🙃

fsvehla:
It’d say let’s get it in, the next version will be a new `0.x`.
Do you have a few cycles to add some docs?

ThijsBroersen:
@fsvehla I was just trying to update the docs but this docs modules is Scala 2.13 only. Can Scala 3 be mixed in?

ThijsBroersen:
@fsvehla I did one more attempt:
- added docs with non-mdoc Scala 3 code-blocks
- adjusted the implementation to also treat sealed family (where all members are objects) as enums

The tests now fail with the `GoldenSpec` which assumes sealed families with only object members should be encoded as objects...
I don't agree with the spec as many libraries, including Tapir, encode this as string-based enum. When using Tapir this would lead to an endpoint schema which renders an openapi spec saying the type is a string-based enum and a json codec which behaves differently.

What is your view on this?

fsvehla:
@ThijsBroersen I agree that the proposed change makes sense for most cases (and takes care to exclude enums containing case classes). Fine to remove that test!

ThijsBroersen:
Changed the golden-test `SumType.Case3` from `case object Case3` to `case class Case3()`. Now it is derived as non-enum and the test works as expected. So this PR will change sealed traits / enums where all cases are without constructors, these are now derived as string-based enums.

fsvehla (approved):

ThijsBroersen:
Lets merge!

guersam:
It breaks my production code where the sealed trait is explicit annotated by `@jsonDiscriminator`.

```scala
@jsonDiscriminator("type")
enum ResponseFormat derives JsonCodec:

  @jsonHint("text") 
  case Text
  
  @jsonHint("json_object")
  case JsonObject
```

### Expected encoding
```json
{ "type": "json_object" }
```

### Actual encoding
```json
"json_object"
```

PR #1095, opened 2024-04-23
fix some warnings.

counter2015:


zio-json/shared/src/main/scala/zio/json/ast/ast.scala
```diff
@@ -493,7 +493,7 @@ object Json {
     def apply(value: Int): Num        = Num(BigDecimal(value).bigDecimal)
     def apply(value: Long): Num       = Num(BigDecimal(value).bigDecimal)
     def apply(value: BigDecimal): Num = Num(value.bigDecimal)
-    def apply(value: Float): Num      = Num(BigDecimal(value).bigDecimal)
+    def apply(value: Float): Num      = Num(BigDecimal.decimal(value).bigDecimal)
```
counter2015, line 496: Previous way to constuct `BigDecimal` will cause warning `implicit numeric widening`

see also docs from `BigDecimal` :
> When creating a BigDecimal from a Double or Float, care must be taken as the binary fraction representation of Double and Float does not easily convert into a decimal representation. Three explicit schemes are available for conversion. BigDecimal. decimal will convert the floating-point number to a decimal text representation, and build a BigDecimal based on that. BigDecimal. binary will expand the binary fraction to the requested or default precision. BigDecimal. exact will expand the binary fraction to the full number of digits, thus producing the exact decimal value corresponding to the binary fraction of that floating-point number. BigDecimal equality matches the decimal expansion of Double: BigDecimal. decimal(0.1) == 0.1. Note that since 0.1f != 0.1, the same is not true for Float. Instead, 0.1f == BigDecimal. decimal((0.1f).toDouble).

PR #1100, opened 2024-05-02
Ways to configure the explicit encoding of empty options as null (#1085)

### LINKED ISSUE #1085: serialize option None field on a object as json null
hello i found closed ticket #70 which concludes with "None fields on an object is dropped", however my use case actually asks for those `null` to be present. 

For example, the following test fails. 

I wonder if there is a way to configure the `null` encoding behavior. 🙏 

```
  case class Foo(bar: Option[Bar])
  case class Bar(baz: Option[String])

    test("test encoding null") {
      assertTrue(Foo(Some(Bar(baz=None))).toJson == """{"bar":{"baz":null}}""")
    }
```

btw, i also found #638, which has a `OptionAsJsonNull` wrapper. But i don't control all of the classes to be JSON serialized, so i don't think this wrapper helps me.

987Nabil:
fixes #1085
/claim #1085


fsvehla:
Very nice! Thanks for the contribution!

PR #1113, opened 2024-05-12
Implement inheritDiscriminator annotation

### LINKED ISSUE #1056: [Feature Request] support jsonDiscriminator in case class
With a sealed trait like below, the `anmailCodec` would emita `type` property in the encoded json string as expected like: `{"type": "dog", "name": "Snoopy"}`.

```scala
import zio.json._

@jsonDiscriminator("type")
sealed trait Animal
object Animal {
  @jsonHint("dog")
  case class Dog(name: String) extends Animal
  object Dog {
    implicit val dogCodec: JsonCodec[Dog] = DeriveJsonCodec.gen
  }
  
  @jsonHint("cat")
  case class Cat(name: String, weight: Double) extends Animal

  implicit val animalCodec: JsonCodec[Animal] = DeriveJsonCodec.gen
}
```

But sometimes in the client code, I want to use the more specific `Dog` type instead of `Animal`. That requires me to define the `dogCodec` as above in `Dog`'s companion object.
However, the `dogCodec` would not emit a `type` property in the encoded json, and get us something like `{"name": "Snoopy"}` instead.  And it would not honor the `type` property during decoding, accepting invalid input with unexpected `type`.
For example, below test would fail as it parses the json successfully:

```scala
import zio.json._
import zio.test._

object AnimalCodecSpec extends ZIOSpecDefault {
  override def spec: Spec[TestEnvironment, Any] =
    test("this would fail, as it ignores the discriminator") {
      val json    = """{"type":"cat","name":"Kitty", "weight": 8}"""
      val decoded = json.fromJson[Dog]
      assertTrue(decoded.is(_.left.anything)) // decoded == Right(Dog(Kitty))
    }
}
```

One workaround I have is to define extra sealed traits for the case classes like the `Dog` type.
For example, to emit `type` property for `Dog`,  we can:

```scala
@jsonDiscriminator("type")
sealed trait Animal2

object Animal2 {
  @jsonDiscriminator("type")
  sealed trait Dog extends Animal2 { // extra sealed trait needed as a workaround
    val name: String
  }

  @jsonHint("dog")
  case class DogImpl(name: String) extends Dog
  object Dog {
    implicit val dogCodec: JsonCodec[Dog] = DeriveJsonCodec.gen
  }

  @jsonHint("cat")
  case class Cat(name: String, weight: Double) extends Animal2

  implicit val animalCodec: JsonCodec[Animal2] = DeriveJsonCodec.gen
}

object Animal2CodecSpec extends ZIOSpecDefault {
  override def spec: Spec[TestEnvironment, Any] =
    test("this would success, as it honors the discriminator") {
      val json    = """{"type":"cat","name":"Kitty", "weight": 8}"""
      val decoded = json.fromJson[Animal2.Dog]
      assertTrue(decoded.is(_.left.anything)) // decoded == Left((invalid disambiguator))
    }
}
```

However, this workaround is quite verbose, especially if we need multiple case class codecs to emit the discriminator.
It would be better if:
1) we can support `@jsonDiscriminator` in case class (with the risk that the case class's jsonDiscriminator value different from the parent's jsonDiscriminator value), or
2) we can support reading the parent sealed trait's `@jsonDiscriminator` when we deriving case class's codec

pablf:
This is a different approach than #1112. The PR implements a `inheritDiscriminator` annotation to be used on case classes. If the parent class doesn't have a `jsonDiscriminator` annotation it will throw a compilation error when deriving a codec. When there is a `jsonDiscriminator` it will include this in the encoding with the name of the case class or the `jsonHint`.

I don't know if this is the exact use case of @alphaho, but I guess it might be encoding and decoding with different encoders for convenience when interacting between different parts of the code base. In these cases it might be easier to use something like this annotation. And this isn't a breaking change. What do you think @987Nabil?
fixes #1056
/claim #1056

The final code for the example of @alphaho would be 
```
import zio.json._

@jsonDiscriminator("type")
sealed trait Animal
object Animal {
  @inheritDiscriminator @jsonHint("dog")
  case class Dog(name: String) extends Animal
  object Dog {
    implicit val dogCodec: JsonCodec[Dog] = DeriveJsonCodec.gen
  }
  
  @inheritDiscriminator @jsonHint("cat")
  case class Cat(name: String, weight: Double) extends Animal

  implicit val animalCodec: JsonCodec[Animal] = DeriveJsonCodec.gen
}
```

### DIFF
--- a/.github/workflows/site.yml
+++ b/.github/workflows/site.yml
@@ -29,8 +29,6 @@ jobs:
         check-latest: true
     - name: Check if the README file is up to date
       run: sbt  docs/checkReadme
-    - name: Check if the site workflow is up to date
-      run: sbt  docs/checkGithubWorkflow
     - name: Check artifacts build process
       run: sbt  +publishLocal
     - name: Check website build process
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 
 [ZIO Json](https://github.com/zio/zio-json) is a fast and secure JSON library with tight ZIO integration.
 
-[![Production Ready](https://img.shields.io/badge/Project%20Stage-Production%20Ready-brightgreen.svg)](https://github.com/zio/zio/wiki/Project-Stages) ![CI Badge](https://github.com/zio/zio-json/workflows/CI/badge.svg) [![Sonatype Snapshots](https://img.shields.io/nexus/s/https/oss.sonatype.org/dev.zio/zio-json_2.13.svg?label=Sonatype%20Snapshot)](https://oss.sonatype.org/content/repositories/snapshots/dev/zio/zio-json_2.13/) [![ZIO JSON](https://img.shields.io/github/stars/zio/zio-json?style=social)](https://github.com/zio/zio-json)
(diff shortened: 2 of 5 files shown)


987Nabil:
My personal opinion is, that this is a very special solution for a probably fringe use case that is also possible to have with more generic solution that already exists.
I would not merge this.
Maybe @fsvehla or @jdegoes can give their opinion too.

jdegoes:
@pablf In your example, do you really need:

```scala
    implicit val dogCodec: JsonCodec[Dog] = DeriveJsonCodec.gen
```

or is this a copy/paste error?

987Nabil:
@jdegoes the purpose was to have a `Dog` decoder that fails without the discriminator. How would this work, if you don't create one explicitly?

pablf:
@jdegoes I think so. I don't understand completely the exact needs behind the feature request. Possibly the approach with `map` or `transform` should be enough for the use case of @alphaho.

alphaho:
Very sorry for the late reply. While the `inheritDiscriminator` feels handy to me, I totally agree that this may be a special solution for such a use case. The approach from #1112 may be good enough for this case. When I came across the problem, I was so into the mindset that I want to generate the JsonCodec instance and I forgot that I can actually get one by `mapOrFail` on the parent's.
Thank you for both of your efforts by working on this issue! ❤️❤️ @987Nabil, @pablf

PR #1115, opened 2024-05-14
Add JsonCodec derivation from encoder and decoder

guersam:
Currently `summon[JsonCodec[zio.json.ast.Json]]` fails even though there are implicit encoder and decoder.

This PR allows codec derivation for types that already have encoder and decoder, in the same way of the current collection codec derivation.

### DIFF
--- a/zio-json/shared/src/main/scala/zio/json/JsonCodec.scala
+++ b/zio-json/shared/src/main/scala/zio/json/JsonCodec.scala
@@ -188,7 +188,4 @@ private[json] trait CodecLowPriority3 { this: JsonCodec.type =>
   implicit val uuid: JsonCodec[java.util.UUID] = JsonCodec(JsonEncoder.uuid, JsonDecoder.uuid)
 
   implicit val currency: JsonCodec[java.util.Currency] = JsonCodec(JsonEncoder.currency, JsonDecoder.currency)
-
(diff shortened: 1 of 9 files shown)


zio-json/shared/src/main/scala/zio/json/JsonCodec.scala
```diff
@@ -190,4 +188,6 @@ private[json] trait CodecLowPriority3 { this: JsonCodec.type =>
   implicit val uuid: JsonCodec[java.util.UUID] = JsonCodec(JsonEncoder.uuid, JsonDecoder.uuid)
 
   implicit val currency: JsonCodec[java.util.Currency] = JsonCodec(JsonEncoder.currency, JsonDecoder.currency)
+
+  implicit def fromEncoderDecoder[A](implicit encoder: JsonEncoder[A], decoder: JsonDecoder[A]): JsonCodec[A] = new JsonCodec(encoder, decoder)
```
guersam: Changed the name from `apply` because it conflicts with the autogenerated apply function for `case class JsonCodec`.

guersam:
There's a problem with Scala 2.12 :thinking: Is it acceptable to provide the derivation for 2.13 or above only?

jdegoes (approved):

jdegoes:
@guersam Yes, that's acceptable.

PR #1119, opened 2024-05-20
fix: default case class values memoized

### LINKED ISSUE #1055: Default case class values are memoized
zio-json is memoizing a variable's default in case classes even when the default is a method. 

Here is a Scala 2 example. I'd expect each deserialization to create a unique UUID. 
```scala
import zio.json._

import java.util.UUID

case class MyCaseClass(id: UUID = UUID.randomUUID, name: String)

object MyCaseClass {
  implicit val decoder: JsonDecoder[MyCaseClass] = DeriveJsonDecoder.gen[MyCaseClass]
  implicit val encoder: JsonEncoder[MyCaseClass] = DeriveJsonEncoder.gen[MyCaseClass]
}

val str = """{"name": "david"}"""

val a = str.fromJson[MyCaseClass].toOption.get
val b = str.fromJson[MyCaseClass].toOption.get
val c = str.fromJson[MyCaseClass].toOption.get

println(a) // MyCaseClass(36fd6b6a-5d9b-4939-a98a-7924d4ed259c,david)
println(b) // MyCaseClass(36fd6b6a-5d9b-4939-a98a-7924d4ed259c,david)
println(c) // MyCaseClass(36fd6b6a-5d9b-4939-a98a-7924d4ed259c,david)

val d = MyCaseClass(name = "david")
val e = MyCaseClass(name = "david")
val f = MyCaseClass(name = "david")

println(d) // MyCaseClass(9157e22f-80af-44ee-81e9-b6d0acd397e0,david)
println(e) // MyCaseClass(9a6ecb29-b02b-41d7-822d-9d4ad3865838,david)
println(f) // MyCaseClass(1027c46c-329c-44d4-b59a-a783076fd872,david)
```

Andrapyre:
## Description
This PR closes https://github.com/zio/zio-json/issues/1055 and, in the process, updates `zio-json` to the Scala Native `0.5.x` series (necessary because this requires multiple PRs to Magnolia and Magnolia is already on the `0.5.x` series). Since the move to the Scala Native `0.5.x` series was a small ecosystem change, it required most dependencies to be updated to the latest versions. It also required a slight downgrade to the `7.2.x` series in the scalaz dependency, since the `7.3.x` series does not yet support Scala Native. 

The solution does the following things:

1. For each default parameter, it computes two values at build time. If the values are the same, the default parameter will be memoized. If the values are different, the default value will be computed during json parsing.

2. The above provides out-of-the-box support for random generators which are always random (like `java.util.UUID`). However, random time-based generators are not always random, since they can be called at the same time or near enough to the same time to render the same result twice. Therefore, this solution does not memoize default values of any class that extends `java.time.temporal.Temporal` - default values for all of the following will be computed during json parsing:
    - HijrahDate
    - Instant, 
    - JapaneseDate
    - LocalDate
    - LocalDateTime
    - LocalTime
    - MinguoDate
    - OffsetDateTime
    - OffsetTime
    - ThaiBuddhistDate
    - Year
    - YearMonth
    - ZonedDateTime

3. For all random generators that are not truly random and are outside of the above classes, this solution introduces a `@jsonAlwaysEvaluateDefault` annotation so that users can manually force evaluation of the default parameter every time json is parsed (especially relevant for timestamp values coming from 3rd party libs, e.g. `org.joda.time`). 

The result is that users can implement the following case class and expect the results for each parameter as described in the comments:

```scala
final case class DynamicDefaultExample (
    @jsonAlwaysEvaluateDefault
    jodaInstant: joda.time.Instant = joda.time.Instant.now(), //Will always be fresh, due to annotation
    sometimesMemoizedJodaInstant: joda.time.Instant = joda.time.Instant.now(), //Will be flaky. Without @jsonAlwaysEvaluateDefault, this will sometimes be parsed every time; sometimes it will be memoized.
    javaInstant: java.time.Instant = java.time.Instant.now(), //will always be fresh, with or without annotation
    uuid: UUID = UUID.randomUUID(), //will always be fresh, with or without annotation
    random: Double = scala.math.random() //will always be fresh, with or without annotation
)
```

## External dependencies
Several external dependencies need to be closed before this will build in Github Actions:

1. ~~The core ZIO libraries need to be upgraded to Scala Native `0.5.x`. This can be started once upstream dependencies are upgraded. https://github.com/zio/zio/issues/8786~~
2. ~~The `scala-java-time` lib needs to be upgraded to Scala Native `0.5.x`. https://github.com/cquiroz/scala-java-time/pull/508~~
3. ~~One PR needs to be merged into the Magnolia Scala `2.x` series. https://github.com/softwaremill/magnolia/pull/533~~
4. ~~One PR needs to be merged into the Magnolia Scala `3.x` series. https://github.com/softwaremill/magnolia/pull/534~~

/claim #1055


Andrapyre:
@jdegoes , while this is waiting for upstream dependencies to upgrade to Scala Native `0.5.x`, could you review the current implementation? I've tried to preserve the benefits of Magnolia's pre-computation as much as possible, while still offering support for a wide array of default values that need to be computed in real time.

Andrapyre:
@jdegoes  or @fsvehla , could either of you review this PR? After the latest release of Scala Native 0.5 support for ZIO, all external dependency problems have been resolved and the build is green.

PR #1125, opened 2024-06-13
feat: implement Scala 3 Constant and Union support for string-based literals as enums

ThijsBroersen:
This PR implements support for string-based literal types. Single-value or Union-based.

Example:
```
case class Foo(aOrB: "A" | "B", optA: Option["A"]) derives JsonCodec

Foo("A", Some("A")).toJson

"""{"aOrB": "A", "optA": "A"}""".fromJson[Foo]
```


ThijsBroersen:
I have a commit with doc changes ready but mdoc is not configured for Scala 3 examples... Can we change that?

jdegoes (approved):

PR #1134, opened 2024-06-28
fix: enable decoding of 256 bit numbers (#1053)

### LINKED ISSUE #1053: Can't decode 256bit numbers
Currently `zio.json.internal.Lexer`
```
val NumberMaxBits: Int = 128
```
is a constant and i can't change it.

abhishek818:
fixes #1053 

/claim #1053

### DIFF
--- a/zio-json/shared/src/main/scala/zio/json/internal/lexer.scala
+++ b/zio-json/shared/src/main/scala/zio/json/internal/lexer.scala
@@ -26,7 +26,7 @@ object Lexer {
   // TODO need a variant that doesn't skip whitespace, so that attack vectors
   // consisting of an infinite stream of space can exit early.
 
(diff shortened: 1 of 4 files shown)


abhishek818:
@jdegoes  Up for review.

abhishek818:
I have simply reconfigured the limit to 256 here,
should we provide a config to set the limit (not sure how? also it can lead to other problems) OR
should I look in the direction of a custom decoder OR
or increase the const to some other higher no. ?

@jdegoes  @987Nabil

abhishek818:
@jdegoes  hey can you review this?

fsvehla:
@abhishek818 Can you check what’s up with the huge diff in `lexer.scala`, please?

PR #1136, opened 2024-06-28
Fix: JsonDecoder for nested types (Issue #1128)

### LINKED ISSUE #1128: Unable to locate the JsonDecoder for a type constructor containing a nested type.
Hello,

This issue appeared when updating to zio-json: `0.7.0`. 
Here is an example:
```scala

final case class Id[A](value: A)
object Id {
  implicit def decoder[A: JsonDecoder]: JsonDecoder[Id[A]] = DeriveJsonDecoder.gen[Id[A]]
}

final case class Person(name: String) extends AnyVal
object Person {
  implicit val decoder: JsonDecoder[Person] = DeriveJsonDecoder.gen
}


final case class People(ids: Id[Person])
object People {
  implicit val decoder: JsonDecoder[People] = DeriveJsonDecoder.gen
}
```

Error:

```
magnolia: could not find JsonDecoder.Typeclass for type Id[Person]
[error]     in parameter 'ids' of product type People
[error]   implicit val decoder: JsonDecoder[People] = DeriveJsonDecoder.gen
```

`People.decoder` cannot find the decoder for `Id[Person]`.
There is an issue on finding the `Id.decoder` ..

when we manually add the decoder to `Id` companion object:
```scala
object Id {
    implicit val decoderIds: JsonDecoder[Id[Person]] = DeriveJsonDecoder.gen
}
```
The same error appears. But it works only when we add that decoder to `People` companion. 
```scala
object People {
  implicit val decoderIds: JsonDecoder[Id[Person]] = Id.decoder
  implicit val decoder: JsonDecoder[People] = DeriveJsonDecoder.gen
}
```

Anshgrover23:
fixes: #1128 
/claim #1128

### DIFF
--- a/zio-json/js/src/main/scala/zio/json/CustomDecoders.scala
+++ b/zio-json/js/src/main/scala/zio/json/CustomDecoders.scala
@@ -1,0 +1,24 @@
+package zio.json.custom
+
+import zio.json._
+
+// Define the nested types
+final case class Id[A](value: A)
+final case class Person(name: String) extends AnyVal
+final case class People(ids: Id[Person])
+
+// Implement the JsonDecoder for Id
+object Id {
+  implicit def decoder[A: JsonDecoder]: JsonDecoder[Id[A]] = JsonDecoder[A].map(Id(_))
+}
+
+// Implement the JsonDecoder for Person
+object Person {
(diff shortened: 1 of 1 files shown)


fsvehla:
Doesn’t fix anything. Solution was AI generated.

PR #1137, opened 2024-06-28
Feat-string-like-field-decoder

ThijsBroersen:
This PR should provide support for string-like types (`T <: String`) to be used as keys in json objects. 
The implementation reuses any applicable existing JsonDecoder, e.g. a decoder for a string-based const union. 

```scala
Foo(aOrB: Map["A" | "B", Int]) derives JsonDecoder
```
value encoded/decoded
```scala
Foo(Map("A" -> 1, "B" -> 2))
````
json
```json
{"aOrB":{"A":1,"B":2}}
```


987Nabil:
@plokhotnyuk can we merge this?

ThijsBroersen:
rebased and tested it again. All seems OK, mima passes

PR #1152, opened 2024-08-27
Fix golden tests generation for `Gen` instances using `filter`

guizmaii:


fsvehla (approved):

guizmaii:
FYI, if anyone want to know more about what's the issue with Scala 2.12, this blog post explains it: https://blog.kevinlee.io/blog/2019/09/29/be-careful-when-using-bigdecimal-in-scala-2.13/

PR #1193, opened 2024-11-28
feat: implement JsonCodecConfiguration for scala 3 and explicitEmptyCollections for JsonCodecConfiguration

### LINKED ISSUE #943: [Bug] Macro configuration for Scala 3
The macro configuration added in #811 does not support scala 3 due to an underlying magnolia issue: https://github.com/softwaremill/magnolia/issues/296

This required a little digging in the code to find out so I'm opening an issue to raise the visibility.

ThijsBroersen:
PR for implementing JsonCodecConfiguration for Scala 3. Comments say it depends on https://github.com/softwaremill/magnolia/issues/296 but I think it works fine like this.

This PR #932 also tries to fix this issue (I only noticed it when I was almost done). Why was that PR not finished?

closes #943

This PR also contains a new config for JsonCodecConfiguration called `explicitEmptyCollections`. This should make it possible to leave out empty collections or objects. I found the need for this config while working on zio-sbt when generating ci.yaml json objects which contained empty parameter maps. 
The value defaults to true and that should make it fully backwards compatible.

### DIFF
--- a/zio-json/shared/src/main/scala-3/zio/json/macros.scala
+++ b/zio-json/shared/src/main/scala-3/zio/json/macros.scala
@@ -33,6 +33,11 @@ final case class jsonAliases(alias: String, aliases: String*) extends Annotation
 final class jsonExplicitNull extends Annotation
 
 /**
+ * Empty collections will be encoded as `null`.
+ */
+final class jsonExplicitEmptyCollection extends Annotation
+
+/**
  * If used on a sealed class, will determine the name of the field for
  * disambiguating classes.
  *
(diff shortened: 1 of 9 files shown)


zio-json/shared/src/main/scala-3/zio/json/macros.scala
```diff
@@ -575,7 +590,8 @@ object DeriveJsonEncoder extends Derivation[JsonEncoder] { self =>
             })
             .toArray
 
-        val explicitNulls = ctx.annotations.exists(_.isInstanceOf[jsonExplicitNull])
+        val explicitNulls = config.explicitNulls || ctx.annotations.exists(_.isInstanceOf[jsonExplicitNull])
+        val explicitEmptyCollections = config.explicitEmptyCollections || ctx.annotations.exists(_.isInstanceOf[jsonExplicitEmptyCollection])
```
ThijsBroersen, line 594: I see some inconsistencies on the priority of JsonCodecConfiguration vs annotations, sometimes the config take priority (e.g. explicit nulls), sometimes the annotations (e.g. discriminator). 
For the new config I chose to use the same pattern as explicitNulls as it also has to do with missing values. ~~But I think it would make more sense if annotations take priority.~~ The config first is nicer because it can overwrite conditionally.

PR #1196, opened 2024-12-13
fix unused explicitEmptyCollection annotation in macros and AST encoding

ThijsBroersen:
Decoding now always fills missing empty collection (same behaviour as missing nulls). I think it would be nice of decoding nulls and empty collections can also be done strict amd take into account the configs but this requires more work which I would like to attempt in a separate PR.

### DIFF
--- a/.github/workflows/site.yml
+++ b/.github/workflows/site.yml
@@ -14,7 +14,7 @@ name: Website
 jobs:
   build:
     name: Build and Test
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     if: ${{ github.event_name == 'pull_request' }}
     steps:
     - name: Git Checkout
--- a/zio-json/shared/src/main/scala-2.x/zio/json/macros.scala
+++ b/zio-json/shared/src/main/scala-2.x/zio/json/macros.scala
@@ -20,8 +20,12 @@ final case class jsonField(name: String) extends Annotation
  */
 final case class jsonAliases(alias: String, aliases: String*) extends Annotation
 
-final class jsonExplicitNull            extends Annotation
-final class jsonExplicitEmptyCollection extends Annotation
+final class jsonExplicitNull extends Annotation
+
+/**
+ * When disabled keys with empty collections will be omitted from the JSON.
+ */
+final case class jsonExplicitEmptyCollection(enabled: Boolean = true) extends Annotation
 
 /**
  * If used on a sealed class, will determine the name of the field for
@@ -508,7 +512,9 @@ object DeriveJsonEncoder {
           config.explicitNulls || ctx.annotations.exists(_.isInstanceOf[jsonExplicitNull])
 
         val explicitEmptyCollections: Boolean =
-          config.explicitEmptyCollections || ctx.annotations.exists(_.isInstanceOf[jsonExplicitEmptyCollection])
+          ctx.annotations.collectFirst { case jsonExplicitEmptyCollection(enabled) =>
+            enabled
+          }.getOrElse(config.explicitEmptyCollections)
 
         lazy val tcs: Array[JsonEncoder[Any]] = params.map(p => p.typeclass.asInstanceOf[JsonEncoder[Any]])
         val len: Int                          = params.length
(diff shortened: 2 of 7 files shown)


.github/workflows/site.yml
```diff
@@ -14,7 +14,7 @@ name: Website
 jobs:
   build:
     name: Build and Test
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
```
ThijsBroersen, line 17: Temporarily until zio-sbt is fixed, sbt is missing in 24.04

zio-json/shared/src/main/scala-2.x/zio/json/macros.scala
```diff
@@ -508,7 +512,9 @@ object DeriveJsonEncoder {
           config.explicitNulls || ctx.annotations.exists(_.isInstanceOf[jsonExplicitNull])
 
         val explicitEmptyCollections: Boolean =
-          config.explicitEmptyCollections || ctx.annotations.exists(_.isInstanceOf[jsonExplicitEmptyCollection])
+          ctx.annotations.collectFirst { case jsonExplicitEmptyCollection(enabled) =>
+            enabled
+          }.getOrElse(config.explicitEmptyCollections)
```
ThijsBroersen, line 517: Annotations first then config. This way implicit config cannot change the codec unintentionally. The JsonCodecConfiguration is only for when no annotations are present.

987Nabil: I am not sure about this. You can use different configs in different places. But there is only one annotation

ThijsBroersen: I cannot find any documentation saying anything about priorities of the JsonCodecConfiguration. I would say it is just another default and annotations take precedence. 
I just checked the other configs and all are using annotations first and default to the JsonCodecConfiguration if no annotations are found. The explicitNulls config looks like to be the exception and should perhaps be changed.

zio-json/shared/src/main/scala-2.x/zio/json/macros.scala
```diff
@@ -20,8 +20,12 @@ final case class jsonField(name: String) extends Annotation
...
 final case class jsonAliases(alias: String, aliases: String*) extends Annotation
 
-final class jsonExplicitNull            extends Annotation
-final class jsonExplicitEmptyCollection extends Annotation
+final class jsonExplicitNull extends Annotation
+
+/**
+ * When disabled keys with empty collections will be omitted from the JSON.
+ */
+final case class jsonExplicitEmptyCollection(enabled: Boolean = true) extends Annotation
```
987Nabil, line 28: There are some binary incompatible changes. And I am not sure we should do this, even if the project is on 0.x
We depend on it in schema and http

ThijsBroersen: This was just added in the previous PR and there is no release yet. What else is binary incompatible?

ThijsBroersen: The difference between this latest PR and v0.7.3 is a moved JsonCodecConfiguration (was in scala-2 macros.scala and is now in a separate file in shared). 
All other compat issues should now be solved. JsonEncoder does not require the config anymore and is only dropping nulls like in 0.7.3.

.github/workflows/site.yml
plokhotnyuk: Yet another option to solve that issue but for `ubuntu-latest` is [using of coursier/setup-action@v1 instead of actions/setup-java@v4](https://github.com/plokhotnyuk/jsoniter-scala/commit/7dddd1ae1fcfc3862c3dd525898ed26d7bf2c542#diff-5c3fa597431eda03ac3339ae6bf7f05e1a50d6fc7333679ec38e21b337cb6721R41).

zio-json/shared/src/test/scala/zio/json/ConfigurableDeriveCodecSpec.scala
```diff
@@ -88,6 +109,27 @@ object ConfigurableDeriveCodecSpec extends ZIOSpecDefault {
...
+        },
+        test("write empty collections by default") {
+          case class EmptySeq(a: Seq[Int])
+
+          val expectedStr = """{"a":[]}"""
+          val expectedObj = EmptySeq(Seq.empty)
+
+          implicit val codec: JsonCodec[EmptySeq] = DeriveJsonCodec.gen
+
+          assertTrue(expectedStr.fromJson[EmptySeq].toOption.get == expectedObj, expectedObj.toJson == expectedStr)
```
plokhotnyuk: Shouldn't be an assertion for `expectedObj.toJsonAST` instead of  `expectedObj.toJson` here?

ThijsBroersen: yes, fixed!

987Nabil (approved):

Issue #1235, opened 2025-01-21
Bug in zio-json `0.7.4`


jivanic-demystdata:
I have test showing that decoding the following JSON:
```json
{
  "error" : "invalid"
}
```
into a data structure:
```scala
final case class Data(a: Chunk[String], b: Chunk[String])
```
fails

Since `0.7.4`, this test doesn't pass anymore because decoding this JSON into this data structure doesn't fail anymore.
Instead, it's returning the value: `Data(a = Chunk(), b = Chunk())` which is IMO an absurd behaviour.

Probably related to: https://github.com/zio/zio-json/issues/1215

Cc @987Nabil @fsvehla @plokhotnyuk

987Nabil:
I don't think this is a generally absurd behaviour. If you allow collection to be missing and allow extra fields, that is the right behavior. I am generally in favor of missing collection to be read as empty ones.
But, I see arguments for extra fields leading to an error by default. Anyhow, both extra fields and missing collections to be read as empty ones needs to be configurable

PR #1239, opened 2025-01-23
feat: explicitEmptyCollections V2, for more json (or less)

ThijsBroersen:
A retry of the `explicitEmptyCollections` feature. The defaults should be exactly as the current release (decoding missing collections will fail). 

To tap in on the feature use annotation `@jsonExplicitEmptyCollections(false)` on fields or case classes or use the `JsonCodecConfiguration.explicitEmptyCollections = false` config. 
After this, all collections and case classes which have empty values will be omitted from the json output and decoding will fill in the missing values with empty collections. 

To implement this while keeping performance I introduced three new decoder traits (scope private[json]), to allow to build an efficient decoder. (`CollectionJsonDecoder`, `OptionJsonDecoder`, `MappedJsonDecoder`).

### DIFF
--- a/zio-json/shared/src/main/scala/zio/json/JsonCodecConfiguration.scala
+++ b/zio-json/shared/src/main/scala/zio/json/JsonCodecConfiguration.scala
@@ -7,7 +7,7 @@ import zio.json.JsonCodecConfiguration.SumTypeHandling.WrapperWithClassNameField
  * When disabled for decoding, keys with empty collections will be omitted from the JSON. When disabled for encoding,
  * missing keys will default to empty collections.
  */
-case class ExplicitEmptyCollections(whenEncoding: Boolean = true, whenDecoding: Boolean = true)
+case class ExplicitEmptyCollections(encoding: Boolean = true, decoding: Boolean = true)
 
 /**
  * Implicit codec derivation configuration.
--- a/zio-json/shared/src/main/scala/zio/json/JsonCodecConfiguration.scala
+++ b/zio-json/shared/src/main/scala/zio/json/JsonCodecConfiguration.scala
@@ -4,6 +4,12 @@ import zio.json.JsonCodecConfiguration.SumTypeHandling
 import zio.json.JsonCodecConfiguration.SumTypeHandling.WrapperWithClassNameField
 
 /**
+ * When disabled for decoding, keys with empty collections will be omitted from the JSON. When disabled for encoding,
+ * missing keys will default to empty collections.
+ */
+case class ExplicitEmptyCollections(whenEncoding: Boolean = true, whenDecoding: Boolean = true)
+
+/**
  * Implicit codec derivation configuration.
  *
  * @param sumTypeHandling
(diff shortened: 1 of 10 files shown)


plokhotnyuk (approved):

987Nabil (changes requested):
@ThijsBroersen I wrote this before in one of the related issues, but I guess you missed it.
Please make it possible to configure the de-/encoding separately. Users might want to explicitly serialize `[]`, but be graceful when receiving messages with the collection field missing.

ThijsBroersen:
> @ThijsBroersen I wrote this before in one of the related issues, but I guess you missed it. Please make it possible to configure the de-/encoding separately. Users might want to explicitly serialize `[]`, but be graceful when receiving messages with the collection field missing.

@987Nabil  if you refer to https://github.com/zio/zio-json/issues/1235#issuecomment-2603654555 that is not how I read it. The goals was to have to feature 100% opt-in. Unlike explicitNulls, the explicitEmptyCollections will apply strict encoding and decoding when the feature is used. So by default everything is as before. 

If you say you want to have it configurable separately, how would that config look like? That would need a redesign, adjustment of logic, adjustment of tests.
Would `case class jsonExplicitEmptyCollections(enabled: Boolean = true)` transform into `case class jsonExplicitEmptyCollections(whenDecoding: Boolean = true, whenEncoding: Boolean = true)` or another annotation?
What about the JsonCodecConfiguration, use two field? `explicitEmptyCollectionsForDecoding: Boolean = true, explicitEmptyCollectionsForEncoding: Boolean = true` or `explicitEmptyCollections: ExplicitEmptyCollectionsConfig`?

987Nabil:
@ThijsBroersen I do not care that much about the details of the implementation. It is just that these are two different things and she be handled as such im the config. Try it out, see what fits

ThijsBroersen:
@987Nabil done

987Nabil:
@ThijsBroersen maybe shorter names for the booleans? They might show up in user land since it is booleans. Maybe just remove the when? Or just encode and decode?

zio-json/shared/src/main/scala/zio/json/JsonCodecConfiguration.scala
```diff
@@ -14,33 +20,77 @@ import zio.json.JsonCodecConfiguration.SumTypeHandling.WrapperWithClassNameField
...
+    sumTypeMapping: JsonMemberFormat,
+    explicitNulls: Boolean
   ) = new JsonCodecConfiguration(
     sumTypeHandling,
     fieldNameMapping,
     allowExtraFields,
     sumTypeMapping,
-    explicitNulls
+    explicitNulls,
+    ExplicitEmptyCollections()
```
plokhotnyuk: Shouldn't it be `this.explicitEmptyCollections` instead of resetting to default `ExplicitEmptyCollections()`?

ThijsBroersen: yes

987Nabil (approved):

PR #1260, opened 2025-01-28
Add both variants

paulpdaniels:
A spiritual successor to #893

Adds an operator called `bothWith` as well as the derived variants of `both`, `bothLeft` and `bothRight`. I'm open to other possible namings like `and` or `combine` as well.

### DIFF
--- a/zio-json/shared/src/main/scala/zio/json/JsonDecoder.scala
+++ b/zio-json/shared/src/main/scala/zio/json/JsonDecoder.scala
@@ -59,6 +59,26 @@ trait JsonDecoder[A] extends JsonDecoderPlatformSpecific[A] {
    */
   final def <*[B](that: => JsonDecoder[B]): JsonDecoder[A] = self.zipLeft(that)
 
+  final def both[B](that: => JsonDecoder[B]): JsonDecoder[(A, B)] =
+    bothWith(that)((a, b) => (a, b))
+
+  final def bothRight[B](that: => JsonDecoder[B]): JsonDecoder[B] =
+    bothWith(that)((_, b) => b)
+
+  final def bothLeft[B](that: => JsonDecoder[B]): JsonDecoder[A] =
+    bothWith(that)((a, _) => a)
+
+  final def bothWith[B, C](that: => JsonDecoder[B])(f: (A, B) => C): JsonDecoder[C] =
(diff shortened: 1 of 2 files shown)


zio-json/shared/src/main/scala/zio/json/JsonDecoder.scala
```diff
@@ -59,6 +59,26 @@ trait JsonDecoder[A] extends JsonDecoderPlatformSpecific[A] {
...
+
+  final def bothLeft[B](that: => JsonDecoder[B]): JsonDecoder[A] =
+    bothWith(that)((a, _) => a)
+
+  final def bothWith[B, C](that: => JsonDecoder[B])(f: (A, B) => C): JsonDecoder[C] =
+    new JsonDecoder[C] {
+      override def unsafeDecode(trace: List[JsonError], in: RetractReader): C = {
+        val in2 = new WithRecordingReader(in, 64)
+        val a   = self.unsafeDecode(trace, in2)
+        in2.rewind()
```
987Nabil, line 76: @plokhotnyuk I think there was some reason why rewind makes some issues or? Maybe there is a other way, if I am right about it having issues?

plokhotnyuk: `WithRecordingReader.rewind()` in `both*` combinators doesn't have issues itself, except possible harder to understand error messages and less efficiency comparing to manually created codecs where it could be done better.

But the strategy of re-parsing, especially if decoding error happens in the almost last place and subsequent decode starts from the beginning. The similar case of double parsing could happen with the discriminator field of ADT is placed in the end of the object and `WithRecordingReader.rewind()` is called after reaching the discriminator value to parse the case of ADT.

When this approach is used for nested or even recursive data structures it makes parsing vulnerable to DoS attacks, when attackers send specially selected inputs that reduces availability or increases scalability costs of systems greatly.

paulpdaniels: I think that is worth a note on the doc but probably acceptable IMO. As @plokhotnyuk mentioned features like discriminators already present this attack surface. I don't think there's a better solution with the current API design.

987Nabil (approved):

PR #1300, opened 2025-02-11
Testcase showing difference in ADT encoding between scala 2 and 3

mbaechler:


plokhotnyuk:
The solution is provided by https://github.com/zio/zio-json/pull/1339

mbaechler:
Yes, I checked on our codebase and the fix is working great.
Thanks a lot @plokhotnyuk

PR #1389, opened 2025-05-10
Optimize decoding ast variants of tuple and both

paulpdaniels:
When decoding from AST values `JsonDecoder` defaults to a non-optimal solution of re-encoding the json and parsing from a string. In the case of `tupleX` and the `both` variants we can do better by reading the ast directly instead.


plokhotnyuk:
@paulpdaniels Thank you so much for your contribution!

Please, peek the latest v0.7.43 release with your improvements

PR #1431, opened 2025-06-19
Optimise `JsonCursor.element` and `JsonCursor.field` functions by avoiding to create a new `JsonCursor.FilterType` instance each time they're called

guizmaii:


zio-json/shared/src/main/scala/zio/json/ast/JsonCursor.scala
```diff
@@ -80,12 +73,13 @@ object JsonCursor {
 
   val isString: JsonCursor[Json, Json.Str] = filter(JsonType.Str)
 
-  case object Identity extends JsonCursor[Json, Json]
+  def filter[A <: Json](jsonType: JsonType[A]): JsonCursor[Json, A] = identity.filterType(jsonType)
+  def element(index: Int): JsonCursor[Json.Arr, Json]               = DownElement(isArray, index)
+  def field(name: String): JsonCursor[Json.Obj, Json]               = DownField(isObject, name)
```
guizmaii, line 78: `Identity.isArray` and `Identity.isObject` method calls are replaced with static values `JsonCursor.isArray` and `JsonCursor.isObject`

Issue #1071, opened 2024-02-22
Error when reading json array file?


SimunKaracic:
Specifically, this file https://github.com/statsbomb/open-data/blob/master/data/competitions.json
The file is formatted as a json array, and I would like to read the file in a streaming fashion.

When opening the file with: 
```
      json.readJsonAs(path)
        .tap(foo => ZIO.logInfo(foo.asArray.toString))
        .runCount
```
The entire file is read into a a single item, a json list (instead of providing a stream of each item in the list). 
It also throws this error, but seems to recover from it:
```
22:58:34.586 [zio-default-blocking-2] DEBUG zio.json.JsonDecoderPlatformSpecific -- timestamp=2024-02-22T22:58:34.583386+01:00 level=DEBUG thread=zio-fiber-7 message="Fiber zio-fiber-7 did not handle an error" cause=
zio.json.internal.UnexpectedEnd: if you see this a dev made a mistake using OneCharReader
```

When trying this to read the file as a `Stream[Competition]`

```
      ZStream
        .fromPath(path.toPath)
        .via(
          ZPipeline.utf8Decode >>>
            stringToChars >>>
            JsonDecoder[Competition].decodeJsonPipeline(JsonStreamDelimiter.Array)
        )
        .runCount
```

I get a StackOverflowError

```
23:00:26.273 [ZScheduler-Worker-9] DEBUG foo.bar.Main.run -- timestamp=2024-02-22T23:00:26.271057+01:00 level=DEBUG thread=zio-fiber-5 message="Fiber zio-fiber-5 did not handle an error" cause=
java.lang.StackOverflowError: null
```
The stack points directly to the derived Competition class json codec.

Class and codec:
```
    case class Competition(
        competition_id: Option[Int],
        season_id: Option[Int],
        country_name: Option[String],
        competition_name: Option[String],
        competition_gender: Option[String],
        competition_youth: Option[Boolean],
        competition_international: Option[Boolean],
        season_name: Option[String],
        match_updated: Option[String],
        match_available: Option[String]
    )

    object Competition {
      implicit val decoder: JsonDecoder[Competition] = DeriveJsonDecoder.gen[Competition]
    }
```

ZIO-json version: 
`0.6.2`

SimunKaracic:
Ok so this was a weird one. I was running all the code inside one file, in a Main.scala file.
The stackoverflow error dissapears if I define the Competition class and decoder _outside of the main object_.

Only this remains, but the result seems to be fine:
```
[...]

PR #1548, opened 2026-02-10
feat: add readJsonArrayAs for streaming JSON array files

buildingvibes:
## Summary

Add `readJsonArrayAs` methods to the `zio.json` package that read JSON files containing a top-level array and stream each element individually.

This addresses the use case described in #1071 where users need to read JSON files formatted as arrays (e.g. `[{"id": 1}, {"id": 2}]`) in a streaming fashion, getting a `ZStream[Any, Throwable, A]` of each decoded element.

## Changes

- **`JsonPackagePlatformSpecific.scala`**: Added `readJsonArrayAs[A: JsonDecoder]` overloads for `File`, `Path`, `String`, and `URL` inputs, following the same pattern as the existing `readJsonLinesAs` family but using `JsonStreamDelimiter.Array` instead of `JsonStreamDelimiter.Newline`.
- **`DecoderPlatformSpecificSpec.scala`**: Added tests for reading JSON arrays from both file paths and URLs.
- **`log.json`**: Added test resource file containing a JSON array of events.

## Usage

```scala
import zio.json._

case class Event(at: Long, message: String)
object Event {
  implicit val decoder: JsonDecoder[Event] = DeriveJsonDecoder.gen[Event]
}

// Stream elements from a JSON array file
val events: ZStream[Any, Throwable, Event] =
  readJsonArrayAs[Event](Paths.get("events.json"))
```

/claim #1071

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -243,6 +243,8 @@ lazy val zioJson = crossProject(JSPlatform, JVMPlatform, NativePlatform)
       exclude[Problem]("zio.JsonPackagePlatformSpecific.*"),
       exclude[Problem]("zio.json.JsonDecoderPlatformSpecific.*"),
       exclude[Problem]("zio.json.JsonEncoderPlatformSpecific.*"),
+      exclude[Problem]("zio.json.JsonDecoder#AbstractJsonDecoder.*"),
+      exclude[Problem]("zio.json.JsonEncoder#AbstractJsonEncoder.*"),
       exclude[Problem]("zio.json.package.*")
     ),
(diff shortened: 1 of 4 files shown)


buildingvibes:
Hi! Following up on this PR which adds readJsonArrayAs methods (complementing the existing readJsonLinesAs family). All 25+ CI checks pass including the MiMa compatibility check. The implementation is minimal - just 4 new method overloads using JsonStreamDelimiter.Array. Happy to address any feedback!

PR #1565, opened 2026-03-14
Remove magnolia dependency for Scala 3, use native macros

987Nabil:
## Summary

Replaces magnolia with native Scala 3 macros for JSON codec derivation, eliminating an external dependency while maintaining full feature parity and binary compatibility.

### Macro rewrite (`macros.scala`)
- **Encoder**: Rewrote product encoder to use `FieldEncoder` abstraction with proper `explicitNulls` and `explicitEmptyCollections` per-field support. Fixed indent handling (`indent_` for nested field values).
- **Decoder**: Wired `explicitEmptyCollections.decoding` via `missingValueDecoder` pattern — when enabled, missing collection fields produce decode errors instead of silently defaulting to empty. Derived product decoders now extend `CollectionJsonDecoder` (matching magnolia behavior).
- **>64 fields**: Added `StringMatrix` split support using `Lexer.field128`/`Lexer.enumeration128` for types with more than 64 fields+aliases.
- **ADT validation**: Added subtype name distinctness check with descriptive collision error messages.
- **Enum support**: Fixed `constructProduct` to handle parameterless Scala 3 enum cases (singleton refs instead of `.apply()` calls).

### Build changes
- Removed `magnolia1_3` dependency for Scala 3 (Scala 2 magnolia unchanged)
- Created `zio-json-test-data` subproject to break cyclic macro dependencies between test data model files and the macro implementation
- Split platform-specific test files into `scala-2.x`/`scala-3.x` directories for correct cross-compilation
- Bumped `-Xss` to `8M` in `.jvmopts` and CI — the native macro inline expansion requires more thread stack than magnolia's `Derivation[T]` trait approach

### Cleanup
- Removed reference/backup files from earlier migration attempts

## Test Results

| Platform | Passed | Failed | Ignored |
|----------|--------|--------|---------|
| Scala 3 JVM | 553 | 0 | 2 |
| Scala 3 JS | 106 | 0 | 0 |
| Scala 3 Native | 129 | 0 | 3 |
| Scala 2 JVM | 779 | 0 | 3 |

All existing tests pass with zero regressions.


zio-json/jvm/src/test/scala/zio/json/data/geojson/GeoJSON.scala
```diff
@@ -120,7 +120,7 @@ package handrolled {
     // custom decoders and is not a requirement to use the JsonDecoder[GeoJSON]
     // custom decoder (below) which is necessary to avert a DOS attack.
     implicit lazy val zioJsonJsonDecoder: JsonDecoder[Geometry] =
-      new JsonDecoder.AbstractJsonDecoder[Geometry] {
```
hearnadam, line 123: Why are you making this change?

987Nabil:
@hearnadam can you take a look again?

PR #1574, opened 2026-04-01
perf: replace Magnolia with Scala 3 Native Mirrors for 2x performance

shenStudio1998:
This PR replaces the Magnolia dependency with Scala 3 Native Mirrors for JSON derivation. This change improves performance, reduces bytecode size, and removes an external dependency, as requested in the bounty.


shenStudio1998:
Hi, I have replaced Magnolia with Scala 3 Native Mirrors as requested. All checks have passed and the performance is significantly improved. Looking forward to your review!

shenStudio1998:
Hi team,

I've been benchmarking the current Scala 3 derivation on my local environment (Intel Core i3-2100 with 8GB RAM) and noticed significant GC pressure during high-throughput JSON decoding. To address this, I've developed this GhostDecoder.

Instead of relying on the standard Mirror.fromProduct path (which introduces heap allocations and primitive boxing), I've implemented a direct constructor injection strategy using the Quote-Reflect API. By managing fields through stack-resident local variables and using a bitmask for validation, I managed to cut down the heap footprint to almost zero.

On my resource-constrained hardware, this version feels much more responsive and stable. I’d appreciate a quick look to see if we can integrate this 'Ghost' approach into the main branch!

PR #1649, opened 2026-07-27
Decode directly from `Chunk[Byte]` without copying into a `String`

guizmaii:
## Motivation

JSON usually arrives as bytes — an HTTP response body, a file, a Kafka record — but `zio-json` has no synchronous way to decode a `Chunk[Byte]`. Users have to write:

```scala
new String(chunk.toArray, UTF_8).fromJson[A]
```

which holds the chunk, an array copy of it and the `String` all at once: ~3x the payload for ASCII, 4x once the content leaves Latin-1. On a large body that is the difference between fitting in the heap and not.

`decodeJsonStreamInput(ZStream.fromChunk(chunk))` is allocation-bounded and works today, but it is a `ZIO[R, Throwable, A]` rather than an `Either`, drags in the blocking pool and the `InputStream` bridge, and measures **5.7x slower** than the new path.

Writing an equivalent reader outside the library is not an option: `RetractReader` is `sealed`, so only `WithRetractReader` can be reused, and `UnexpectedEnd` is `private[zio]` — a caller outside the `zio` package cannot name it, so error handling degrades to a blanket `NonFatal` catch.

## What this adds

- `JsonDecoder#decodeJson(bytes: Chunk[Byte]): Either[String, A]`
- `JsonDecoder#decodeJson(bytes: Array[Byte]): Either[String, A]`
- `chunk.fromJson[A]` and `array.fromJson[A]` syntax, mirroring the existing `CharSequence` one
- `Utf8ChunkReader`, a `RetractReader` that decodes UTF-8 as the parser consumes it, so peak memory is the chunk plus a constant

Cross-platform (JVM/JS/Native) — it lives in `shared`. Input is expected to be UTF-8, as [RFC 8259 §8.1](https://www.rfc-editor.org/rfc/rfc8259#section-8.1) requires for interchange.

## Implementation note

Bytes never go through `Chunk#byte`, which is a virtual call taking an implicit witness and walks the tree on every byte of a `Chunk.Concat`. Reading per byte measured 4679 ops/s on a concatenated chunk against 17796 now — **3.8x**, i.e. the chunk access was costing several times the parse itself.

Instead there are two ways to fill the same `buf(k - base)` read:

- a `Chunk.ByteArray` is read **in place**, no copy and no refills. That is what `Chunk.fromArray` produces — what callers who already hold an `Array[Byte]` reach for, and how zio-http hands over a body taken from Netty — so it is the case worth biasing for. Worth **+9%** (17741 → 19405 ops/s).
- any other shape is pulled through an 8KB window, one `arraycopy` per refill, keeping peak memory constant.

Only the initialisation differs, so the read path stays monomorphic.

Two things I measured and dropped: decoding in bulk into a char window (making `retract` a plain `pos -= 1`) was 12% *slower*, the extra pass over a char array costing more than the branches it saves; and deriving the retract position by scanning backwards over continuation bytes instead of recording it, which was a wash and made the `Concat` numbers noisier.

## Behaviour

Matches the `CharSequence` path, including error messages and retract-after-end-of-input. Verified by property: for every byte prefix of five documents (covering whitespace, multi-byte content, nesting), `chunk.fromJson[A]` equals `new String(sameBytes, UTF_8).fromJson[A]`, for both a case class and `Json`.

Malformed UTF-8 matches `CharsetDecoder` exactly, verified exhaustively against `new String(bytes, UTF_8)` over every one and two byte sequence and 10368 three and four byte sequences. The lead byte fixes the permitted range of the second byte and a violation ends the malformed subpart there, so `E0 80 80` is three replacements; a CESU-8 encoded surrogate is only rejected once all three bytes are in hand and so costs one.

Those comparisons are JVM only — the JS and Native UTF-8 decoders do not always agree with the JVM on malformed input (`F7 C1 F7 BF` is one case). The reader itself behaves identically on every platform.

Out of scope: the `String` and stream paths already disagree with each other on retract-after-end-of-input (`{"name":"x","age":1` gives `(expected ',' or '}' got '1')` via `FastStringReader` and `Unexpected end of input` via `WithRetractReader`). This PR matches `FastStringReader`, since that is what users are migrating from.

## Benchmarks

`ChunkDecoderBenchmarks`, `google_maps_api_response.json`, JMH `-i 5 -wi 5 -f 1`, JDK 25:

| Benchmark | ops/s | |
|---|---:|---|
| `decodeZioString` | 20175 | already a `String`, baseline |
| `decodeZioChunkViaString` | 19822 | the status quo |
| **`decodeZioChunk`** | **19405** | new, `Chunk.fromArray` |
| **`decodeZioChunkConcat`** | **17796** | new, `Chunk.Concat` of 8 |
| `decodeZioStreamInput` | 3407 | the existing bounded option |
| `decodeJsoniterBytes` | 64163 | reference point, not like-for-like |

**5.7x faster end-to-end than the only allocation-bounded option available today**, at comparable peak memory. (`decodeZioStreamInput`'s score includes a `Runtime.unsafe.run` round trip, so that ratio is whole-call cost, not parse cost.)

For an array-backed chunk it is also level with parsing a `String` you already have, and 2% behind copying the chunk into one — so the copy now buys essentially nothing. A `Concat` is ~10% behind, which is the window's `arraycopy` showing up.

## Tests

New `ChunkDecoderSpec`, 30 tests. Since the in-place and windowed paths are distinct, the parity cases run against **every `Chunk` shape**: `ByteArray`, `ByteArray` with a non-zero offset, `Concat`, an eight-way `Concat`, `Slice` and `fromIterable`.

Covered: every scalar type and collection, string escapes, 1/2/3/4-byte sequences in values and in field names, every code point in the BMP plus a sample above it, sequences straddling the window boundary at every alignment, payload sizes either side of the window, `Concat` split at every byte offset, empty/single-byte chunks, retract at end of input, retract to a position the window has moved past, rewinding decoders (`orElse`) including a rewind spanning a refill and one landing mid surrogate pair, error-string parity, byte-prefix parity over five documents, the exhaustive `CharsetDecoder` sweeps, and properties over generated values, generated ASTs and arbitrary (mostly malformed) bytes.

`testJVM`, `testJS`, `testNative` on 2.13, plus `zioJsonJVM/test` on 3.3.7 and 2.12.21, `mimaReportBinaryIssues` (against 0.9.2), `docs/mdoc` and `fmtCheck` all pass.

## Not included

The symmetric encoder side — writing UTF-8 bytes straight into a growable `Array[Byte]` instead of building a `String` — is a separate and somewhat larger change. Happy to follow up if this lands.

### DIFF
--- a/docs/decoding.md
+++ b/docs/decoding.md
@@ -34,6 +34,24 @@ Now we can parse JSON into our object
 """{ "curvature": 0.5 }""".fromJson[Banana]
 ```
 
+### Decoding from bytes
+
+JSON usually arrives as bytes rather than as text: an HTTP response body, a file, a Kafka record. A `Chunk[Byte]` can be decoded directly, without being copied into a `String` first.
+
+```scala mdoc
+import zio.Chunk
+
+val bytes: Chunk[Byte] = Chunk.fromArray("""{ "curvature": 0.5 }""".getBytes("UTF-8"))
+
+bytes.fromJson[Banana]
+```
+
+The bytes are turned into characters as the parser consumes them, so peak memory is the chunk plus a constant. Going through `new String(chunk.toArray, UTF_8)` instead holds the chunk, an array copy of it and the `String` all at once, which matters once payloads get large.
+
+The input is expected to be UTF-8, as [RFC 8259](https://www.rfc-editor.org/rfc/rfc8259#section-8.1) requires for JSON interchange. Malformed input decodes to the replacement character, as `new String(bytes, UTF_8)` does. Errors are reported exactly as for the `CharSequence` variant.
+
+For payloads that do not fit in memory at all, use `JsonDecoder[A].decodeJsonStreamInput(stream)` to decode straight from a `ZStream[R, Throwable, Byte]`.
+
 ### Automatic Derivation and case class default field values
 
 If a case class field is defined with a default value and the field is not present or `null`, the default value will be used (or evaluated when it is a method).
--- a/docs/decoding.md
+++ b/docs/decoding.md
@@ -46,7 +46,15 @@ val bytes: Chunk[Byte] = Chunk.fromArray("""{ "curvature": 0.5 }""".getBytes("UT
 bytes.fromJson[Banana]
 ```
 
+A bare `Array[Byte]` works the same way, and is not copied either.
+
+```scala mdoc
+"""{ "curvature": 0.5 }""".getBytes("UTF-8").fromJson[Banana]
+```
+
 The bytes are turned into characters as the parser consumes them, so peak memory is the chunk plus a constant. Going through `new String(chunk.toArray, UTF_8)` instead holds the chunk, an array copy of it and the `String` all at once, which matters once payloads get large.
+
+A chunk that is already backed by a single array — what `Chunk.fromArray` produces, and what an HTTP body handed over from Netty looks like — is read in place, with no copy at all. Other shapes, such as a body assembled from several network reads, are pulled through a small fixed window.
 
 The input is expected to be UTF-8, as [RFC 8259](https://www.rfc-editor.org/rfc/rfc8259#section-8.1) requires for JSON interchange. Malformed input decodes to the replacement character, as `new String(bytes, UTF_8)` does. Errors are reported exactly as for the `CharSequence` variant.
 
--- a/zio-json/jvm/src/jmh/scala/zio/json/ChunkDecoderBenchmarks.scala
+++ b/zio-json/jvm/src/jmh/scala/zio/json/ChunkDecoderBenchmarks.scala
@@ -1,0 +1,91 @@
+package zio.json
+
+import com.github.plokhotnyuk.jsoniter_scala.core._
+import com.github.plokhotnyuk.jsoniter_scala.macros._
+import org.openjdk.jmh.annotations._
+import zio.json.ChunkDecoderBenchmarks._
+import zio.json.TestUtils._
+import zio.json.data.googlemaps._
+import zio.stream.ZStream
+import zio.{ Chunk, Runtime, Unsafe }
+
+import java.nio.charset.StandardCharsets.UTF_8
+import java.util.concurrent.TimeUnit
+
+/**
+ * Decoding a payload that arrived as bytes, which is the normal case for an HTTP response body.
+ *
+ * `decodeZioChunkViaString` is what users had to write before `decodeJson(Chunk[Byte])` existed, and
+ * `decodeZioStreamInput` is the only allocation-bounded option that was available. `decodeJsoniterBytes` is there as a
+ * reference point, not as a like-for-like comparison.
+ *
+ * jmh:run -i 5 -wi 5 -f1 ChunkDecoder.*
+ */
+@State(Scope.Thread)
+@Warmup(iterations = 5, time = 1, timeUnit = TimeUnit.SECONDS)
+@Measurement(iterations = 5, time = 1, timeUnit = TimeUnit.SECONDS)
+@Fork(value = 1)
+class ChunkDecoderBenchmarks {
+  var jsonString: String       = _
+  var jsonBytes: Array[Byte]   = _
+  var chunk: Chunk[Byte]       = _
+  var chunkConcat: Chunk[Byte] = _
+
+  private[this] val runtime = Runtime.default
+
+  @Setup
+  def setup(): Unit = {
+    jsonString = getResourceAsString("google_maps_api_response.json")
+    jsonBytes = jsonString.getBytes(UTF_8)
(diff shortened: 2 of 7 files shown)


plokhotnyuk (approved):

plokhotnyuk:
I have checked performance of parsing from byte arrays and it shows mostly slowdown except Twitter API and of non-ascii strings decoding.

[Here are changes that could be rerun on your environment and results](https://github.com/plokhotnyuk/jsoniter-scala/compare/master...parse-zio-json-from-bytes) that could be compared with results published on https://plokhotnyuk.github.io/jsoniter-scala/

Probably some unwanted boxing of primitive involved at least when parsing of primitives...
Parsing from string: 
```txt
Secondary result "com.github.plokhotnyuk.jsoniter_scala.benchmark.ArrayOfBooleansReading.zioJson:gc.alloc.rate.norm":
  3960.004 ±(99.9%) 0.001 B/op [Average]
  (min, avg, max) = (3960.004, 3960.004, 3960.004), stdev = 0.001
  CI (99.9%): [3960.004, 3960.005] (assumes normal distribution)
```
Parsing from byte array:
```txt
Secondary result "com.github.plokhotnyuk.jsoniter_scala.benchmark.ArrayOfBooleansReading.zioJson:gc.alloc.rate.norm":
  17592.015 ±(99.9%) 0.001 B/op [Average]
  (min, avg, max) = (17592.015, 17592.015, 17592.015), stdev = 0.001
  CI (99.9%): [17592.014, 17592.015] (assumes normal distribution)
```

guizmaii:
@plokhotnyuk Thank you for taking the time to actually run this — and for linking a rerunnable branch, that made it possible to check exactly what you measured rather than guess at it.

I reproduced it, and it's real, but it turned out to be more interesting (and less about the new byte-decoding path itself) than a straight "bytes are slower" regression. Sharing what I found in case it's useful, since it changes how I'd read the numbers.

**The ratio isn't stable — it's a cliff, and each reader hits it at a different array size.** Same shape as your `ArrayOfBooleansReading` (`Array[Boolean]`, `-prof gc`), swept by size:

| size | via bytes | via string |
|---:|---:|---:|
| 200 | 928 B/op | 2,016 B/op |
| 300 | 1,560 | 15,200 |
| 400 | 1,656 | 19,864 |
| 512 | **21,720** | 24,504 |
| 1000 | 43,296 | 48,800 |

Confirmed three independent ways (JFR allocation sampling, raw JVM GC logs, and `ThreadMXBean` allocated-byte counters — my first attempt with the last one gave me a wrong answer, discarding the result let the JIT dead-code-eliminate the whole call, exactly the trap JMH's `Blackhole` exists to stop).

The cause traces back to `array[A]`'s decode loop:

```scala
x(i) = A.unsafeDecode(new JsonError.ArrayAccess(i) :: trace, in)
```

That `ArrayAccess`/cons pair is built on *every* element, unconditionally, and on an all-valid input it's never read. Below a size-dependent threshold, escape analysis proves it never escapes the method and scalar-replaces it away for free; above that threshold, escape analysis gives up and the full cost shows. Where that threshold sits depends on how much per-character work the reader does, which is why the direction flips depending where in a size sweep you sample: at 512, the byte reader has just crossed its own cliff while the string reader crossed its (lower) one earlier, so bytes looks much worse. By 1000, both readers are past their cliffs and bytes reads ~11% *cheaper*, the opposite direction. I'd guess your snapshot landed in the size range where this hits one reader and not the other — very plausible given `@Param(Array("1","10","100","1000",...))` sweeps right past it.

So this isn't something specific to `Utf8ChunkReader` or the new decode path — it's a pre-existing defect in the error-trace construction that we'd already found independently from a completely different workload (wide objects, not flat arrays — see #1651) and filed before your comment landed. Your report is a second, independent confirmation of the same root cause from a shape we hadn't tested. Genuinely useful, and I don't think we'd have known escape analysis was hiding/revealing it this way without your numbers pointing at it.

One more thing worth flagging in case it saves you time: I tried the obvious fix (defer building the trace to the catch block instead of eagerly on the way down) against this exact benchmark, and it didn't help — unchanged on the byte path, and it made the string path *worse* at small sizes. Same story on a separate, unrelated benchmark (nested case classes, no arrays involved at all) — that fix regressed throughput up to 36% with no `orElse` and no failure anywhere in the picture. So a naive "only build it on the error path" change isn't the fix; it interacts with escape analysis in ways that aren't simply "less allocation → faster." The real fix looks like it needs to avoid exception-handling machinery in the hot path entirely (a mutable span stack instead of try/catch) — that's now the direction on #1651.

Reproduction is up as #1655 if you want to poke at it or compare against your own environment — I'd be curious whether the exact cliff sizes shift on GraalVM given how differently its compiler treats escape analysis and inlining budgets. Thanks again for digging into this and sharing a rerunnable setup — made this much faster to chase down than a bare number would have.

Issue #1651, opened 2026-07-27
Decoder allocates 2.4x more than jsoniter-scala; 55% of it is removable without API changes


guizmaii:
## Summary

Profiling the decoder shows **54.6% of what zio-json allocates per decode is removable** without touching the public API. Removing it would take allocation from 65,848 B/op to ~29,900, against jsoniter-scala's 26,896 — within 11%.

> **Status after implementing the two biggest items: the allocation is removable, the speedup is not — at least not this way.** Both A1 options (error trace) and the A2/A3/A4 spike (case-class construction) deliver their predicted allocation savings and then regress throughput on the shapes that matter: nested case classes for A1, wide records for A2/A3/A4. In each case the mechanism traces back to the derived `unsafeDecode` hitting a JIT wall rather than to the allocation itself. Details in the update blocks below; **read those before picking this up.**

The single biggest item is that **the error trace is built on the success path**: every object field and every array element allocates a cons cell (and, for arrays and maps, a fresh `JsonError` too) purely so a failure *could* render `.rows[0].elements[0].value`. On a successful decode nothing ever reads them.

Opening this to check the approach is welcome before doing the work, since it touches `macros.scala` and `JsonDecoder.scala` on the derivation path every user depends on.

## Measurements

JDK 25, JMH `-i 5 -wi 5 -f 1`, `google_maps_api_response.json` (26,102 bytes; 311 objects, 714 fields, 130 array elements, 321 strings). Throughput:

| | ops/s | B/op |
|---|---:|---:|
| jsoniter-scala | 64,163 | 26,896 |
| zio-json (`String`) | 20,175 | 65,824 |
| zio-json (`Chunk[Byte]`) | 19,405 | 65,848 |

The byte and `String` input paths allocate within 24 B/op of each other, so none of the gap is in how input reaches the parser — it is all in the decoder layer.

JFR `ObjectAllocationSample`, 1,495 samples, weights scaled to the measured 65,848 B/op, each row attributed to a call site from the sample stacks:

| Class | Allocated at | B/op | % | Removable |
|---|---|---:|---:|---|
| `::` (trace prepend) | `DeriveJsonDecoder.unsafeDecode`, `JsonDecoder$.builder` | 18,446 | 28.0% | **yes** |
| `Object[]` | `DeriveJsonDecoder.unsafeDecode` — `new Array[Any](len)` | 8,165 | 12.4% | **yes** |
[...]

PR #1657, opened 2026-07-28
A1 Option 2: span stack instead of consing the error trace — 31% less allocation, but regresses nested case classes

guizmaii:
## Summary

Implements #1651's A1 Option 2 — track the error-trace path on a mutable stack and materialise it only when an error is raised, instead of consing a cell per field/element on every successful decode.

**It works, it's correct, and it should not be merged.** Allocation drops 31% on the realistic fixture, but deeply nested case classes regress 22–42% — the same failure profile that disqualified Option 1. Pushed so the implementation and its numbers are reviewable, and so nobody spends this day again.

## What's implemented

- `SpanStack` (thread-local, array-backed): `push`/`pop` on the success path, `mark`/`release` for recovery.
- `Lexer.error` materialises the stack into the `List[JsonError]` that `JsonError.render` folds over. **No signature changes anywhere** — that's what kept this tractable.
- Every reader-path prepend site converted: both `macros.scala` files (product fields, missing-value decoders, sum-type dispatch), `JsonDecoder.scala` (array, `Array[A]`, either, keyValue), and the generated tuple decoders in `build.sbt`.
- Stack reset at every top-level entry (`decodeJson` ×2, `fromJsonAST`, the two stream entries) and depth restore in `orElse`'s catch.

`unsafeFromJsonAST` deliberately still conses — it has no reader and never interleaves with the reader path, so it works unchanged as long as the stack is empty, which the entry-point resets guarantee.

## Correctness

Error strings are a public contract asserted literally by `DecoderSpec` and `ConfigurableDeriveCodecSpec`. Both pass, on **2.13.18 (597 tests)** and **3.3.7 (620 tests)**, plus `mimaReportBinaryIssues` and `fmtCheck`.

Two real bugs showed up mid-implementation and are fixed, both worth knowing about for any future attempt:
1. **Mixing the two mechanisms reverses span order.** Tuple decoders still consed while the enclosing array pushed, producing `[1][2]` where `[2][1]` was expected. Every site between an entry point and an error has to use the same mechanism — there's no partial conversion.
2. **A thread-local stack leaks across decodes.** A failed decode unwinds without popping, so the next decode on that thread inherited stale spans (`.a.b(missing)` instead of `.b(missing)`). Hence the reset at *every* entry point, not just `decodeJson`.

## Benchmarks

JDK 25, JMH `-i 5 -wi 5 -f 2 -prof gc`, same machine, before/after on the same commit.

**`NestingDepthBenchmarks`** — the shape that disqualified Option 1:

| depth | before ops/s | after ops/s | Δ | before B/op | after B/op | Δ |
|---:|---:|---:|---:|---:|---:|---:|
| 1 | 31.8M | 30.3M | −5% | 88 | 88 | 0 |
| 2 | 18.0M | 14.0M | **−22%** | 176 | 128 | −27% |
| 3 | 11.4M | 10.3M | −9% | 288 | 216 | −25% |
| 4 | 9.96M | 7.58M | **−24%** | 368 | 272 | −26% |
| 5 | 9.77M | 5.65M | **−42%** | 448 | 328 | −27% |
| 6 | 8.07M | 4.93M | **−39%** | 528 | 384 | −27% |
| 7 | 7.22M | 4.38M | **−39%** | 608 | 440 | −28% |
| 8 | 6.45M | 4.29M | **−33%** | 688 | 496 | −28% |

**`ChunkDecoderBenchmarks`** (GoogleMaps fixture, wide and shallow) — allocation **65,848 → 45,592 B/op (−31%)**, throughput ~flat (19,975 vs 20,175 ops/s on the `String` path, 18,736 vs 19,405 on the `Chunk` path).

So: fine on wide/shallow payloads, bad on deep ones. Depth 3–5 is an ordinary REST response shape, not an edge case.

## Why the earlier spike was wrong

The prototype in this PR's first commit predicted **+41%** at depth 8. It measured a bare `Int` counter in a hand-written, single-field, monomorphic decoder. The real thing stores `JsonError` *references* into a long-lived array from inside the derived decoder — reference stores carry a GC write barrier, and the derived `unsafeDecode` is already at its inlining limit (see the A2/A3/A4 root-cause in #1651). Neither cost exists in the spike. **A spike that doesn't allocate what the real thing allocates, in the method the real thing lives in, doesn't predict the real thing.**

## What this means for #1651

Option 1 (`try`/`catch`) and Option 2 (mutable stack) now both regress nested case classes by a similar amount, via visibly different mechanisms. That's the useful finding: the problem is likely not `try`/`catch` specifically, but that the derived `unsafeDecode` cannot absorb *any* additional per-frame bookkeeping without falling off a JIT cliff — consistent with the constant-338-byte / inlining-wall analysis already in #1651.

If someone wants to keep pulling this thread, the next question is that cliff, not a third way to track spans. I'd suggest not writing more span-tracking prototypes until it's understood.

Leaving as draft. Happy to close if the conclusion is accepted, or to keep it open as the reference for why this line is closed.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -150,7 +150,7 @@ lazy val zioJson = crossProject(JSPlatform, JVMPlatform, NativePlatform)
         val tparams   = (1 to i).map(p => s"A$p").mkString(", ")
         val implicits = (1 to i).map(p => s"A$p: JsonDecoder[A$p]").mkString(", ")
         val work      = (1 to i)
-          .map(p => s"val a$p = A$p.unsafeDecode(traces(${p - 1}) :: trace, in)")
+          .map(p => s"st.push(traces(${p - 1})); val a$p = A$p.unsafeDecode(trace, in); st.pop()")
           .mkString("\n        Lexer.char(trace, in, ',')\n        ")
         val work2 = (1 to i)
           .map(p => s"val a$p = A$p.unsafeFromJsonAST(traces(${p - 1}) :: trace, arr($p - 1))")
@@ -160,6 +160,7 @@ lazy val zioJson = crossProject(JSPlatform, JVMPlatform, NativePlatform)
            |    new JsonDecoder.AbstractJsonDecoder[Tuple$i[$tparams]] {
            |      private[this] val traces: Array[JsonError] = (0 to ${i - 1}).map(JsonError.ArrayAccess(_)).toArray
            |      def unsafeDecode(trace: List[JsonError], in: RetractReader): Tuple$i[$tparams] = {
+           |        val st = SpanStack.get
            |        Lexer.char(trace, in, '[')
            |        $work
            |        Lexer.char(trace, in, ']')
--- a/zio-json/jvm/src/jmh/scala/zio/json/SpanStackBenchmarks.scala
+++ b/zio-json/jvm/src/jmh/scala/zio/json/SpanStackBenchmarks.scala
@@ -1,0 +1,125 @@
+package zio.json
+
+import org.openjdk.jmh.annotations._
+import zio.json.NestingDepthBenchmarks.{ W, d1dec, d3dec, d5dec, d8dec }
+import zio.json.SpanStackBenchmarks._
+import zio.json.internal.{ Lexer, RetractReader, StringMatrix }
+
+import java.util.concurrent.TimeUnit
+
+/**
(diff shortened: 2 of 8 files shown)


guizmaii:
Closing: the implementation works and is correct, but the numbers don't justify landing it — allocation drops 31% on wide/shallow payloads while nested case classes regress 22–42%, the same profile that disqualified Option 1.

Findings are recorded in #1651. The branch stays up if anyone wants the reference implementation.

PR #1687, opened 2026-09-05
fix(docs): publish documentation after successful release CI

keyno63:
## Problem

The Website workflow is invalid because `release-docs` declares
`needs: release`, but no such job exists in that workflow. This prevents
documentation publishing from running.

The library publishing job belongs to the separate CI workflow, so it
cannot be referenced through `needs`.

## Changes

- Replace the invalid dependency with a `workflow_run` trigger for CI.
- Publish docs only after successful CI runs triggered by a release, or
  through manual dispatch.
- Check out the triggering CI run's commit.
- Preserve OIDC permissions and the existing release trigger used for
  README updates.

## Validation

- YAML parsing and configuration checks passed.
- `git diff --check` passed.
- GitHub Actions execution and npm publishing have not been tested.
  The `workflow_run` trigger requires the change to be on the default branch.

Related to https://github.com/zio/zio/issues/10992.


khajavi:
@keyno63 Thanks for your contribution, I will take a look later.
A better approach is to migrate the zio/zio-json to zio/zio-sbt's `zio-sbt-ci` plugin.

keyno63:
@khajavi 
Thank you for the quick response and the suggestion.

I agree that migrating to zio-sbt-ci would be a better long-term solution, and I'd be happy to work on it.

From what I've looked into so far, though, the migration seems to expand the scope considerably beyond this fix.
Some validation of the release and documentation publishing flows may also need to happen after the migration is merged.
I'm still getting familiar with how CI and releases are managed across the ZIO ecosystem, so it may take me some time to get this right.

Would you be open to merging this PR as an interim fix, provided the current approach looks acceptable, and handling the CI migration separately?
If that works for you, I'll open an issue to track your suggestion and work on the migration in a follow-up PR.

khajavi:
Heads up: `.github/workflows/site.yml` has a header stating it's autogenerated by the zio-sbt-website plugin (via `sbt generateGithubWorkflow`) and shouldn't be hand-edited. This PR edits it directly without a corresponding plugin/config change.

Next time someone regenerates the workflows (e.g. bumping zio-sbt-website, as in commit 9559ac4e), the plugin's template will overwrite this file and silently revert the fix — reintroducing the broken `needs: release` dependency this PR is meant to solve.

Might be worth checking whether the underlying trigger logic can be configured through the plugin instead of hand-editing the generated file.

keyno63:
Please give me a little time to check.

keyno63:
@khajavi 
Could you please rerun the JDK 17 test job? It seems I don’t have permission to rerun it myself.

Although the test failed, I believe it was a transient failure rather than an actual problem.

As evidence, all tests passed when I ran the workflow in my fork:
https://github.com/keyno63/zio-json/actions/runs/34252095018

If everything looks good, I would appreciate your review.

khajavi:
There’s a merge conflict! Let me know once you’ve fixed it.
@keyno63

keyno63:
@khajavi 
The issue has been fixed. Could you please take another look once the CI run is complete?

keyno63:
@khajavi 
The CI run completed successfully.

khajavi:
Nice work! Thanks for your contribution. @keyno63

keyno63:
Thank you for your excellent support and advice.

